PR #5, opened 2020-06-24
Set up linter

genedwards:
Based on this example PR: https://github.com/brettallenyo/step2020/pull/1

Compared to that PR, there are some differences:
 - Removed css validation as too many errors were encountered.
 - Removed html validation since we don't have any html files.
 - Using `prettier` instead of `clang-format` for JS files, since clang-format freaks out and can't handle JSX properly.
 - Detects OS and uses appropriate clang-format binary (I hope this works on Mac and github CI, only tested on cloud shell).
 - Install react and many more eslint plugins to for eslint to handle react/JSX.

Seems to work:
image


genedwards:
@RCAVelez @zoebarth

PR #6, opened 2020-06-24
Map component

Cindyzhang977:
Create Map Component by calling Map Embed API

- hard code route with origin, destination, and waypoints
- call Map Embed API to render map

image


frontend/src/map/Map.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+
+function Map({destinations}) {
+
+  return (
+    <div className="map-container">
+      <iframe width="600"
```
zoebarth, line 7: Should these assets be in a css file rather than inline?

Cindyzhang977: oh yeah, I just copied it from the API doc, but I shall change

RCAVelez: are we following kebab or camel case?

frontend/src/map/Map.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+
+function Map({destinations}) {
+
+  return (
+    <div className="map-container">
+      <iframe width="600" 
+              height="450" 
+              frameBorder="0"
```
RCAVelez, line 9: These can be in mapComponent.css
Everything should be in a mapComponent folder, and files should be {insert}Component.js/.css

Folders and classes should follow kebab or camel case

mecolmg: Some notes on naming patterns I've seen and used in React:
* Components should be concisely named to describe what they are
* Components and their css/scss files should have the same name and casing, just different file endings
* Components should be PascalCase (so in this case `Map.js` and `Map.css` would be correct)
* Anything that isn't a component is typically `camelCase.js`

https://reactjs.org/docs/faq-structure.html

PR #31, opened 2020-06-25
add basic search bar component

zoebarth:
Create very basic search component with no css styling. Styling will be improved and more will be added to this component in future pull requests.

### DIFF
--- a/frontend/src/search/Search.js
+++ b/frontend/src/search/Search.js
@@ -4,16 +4,14 @@ import FormControl from 'react-bootstrap/FormControl';
 import React from 'react';
 
 /** 
-  Creates basic search bar.
+ * Creates basic search bar.
  */
 function Search() {
   return (
-    <div>
-      <Form inline>
-        <FormControl type="text" placeholder="Search" className="mr-sm-2" />
-        <Button variant="secondary">Search</Button>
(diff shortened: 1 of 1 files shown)


frontend/src/search/Search.js
```diff
@@ -0,0 +1,20 @@
...
+import Form from 'react-bootstrap/Form';
+import FormControl from 'react-bootstrap/FormControl';
+import React from 'react';
+
+/** 
+  Creates basic search bar.
+ */
+function Search() {
+  return (
+    <div>
```
Cindyzhang977, line 11: don't think you need <div></div> since Form is already just one component.

frontend/src/search/Search.js
```diff
@@ -0,0 +1,20 @@
+import Button from 'react-bootstrap/Button';
+import Form from 'react-bootstrap/Form';
+import FormControl from 'react-bootstrap/FormControl';
+import React from 'react';
+
+/** 
+  Creates basic search bar.
```
mecolmg, line 7: nit: missing an `*` there.

frontend/src/search/Search.js
mecolmg: +1

PR #32, opened 2020-06-25
Optimize button 

Cindyzhang977:
Front end of button that would optimize the route. 
image

Use text to indicate that route is optimized.
image


frontend/src/buttons/OptimizeButton.js
```diff
@@ -0,0 +1,30 @@
...
+import Button from 'react-bootstrap/Button';
+import './OptimizeButton.css';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faCheckCircle } from '@fortawesome/free-solid-svg-icons';
+
+/**
+ * Render a button to optimize route on click. 
+ * If route is optimized, render text indicating that it is optimized.
+ */
+function OptimizeButton() {
```
mecolmg, line 11: You may want a prop to set the initial value of isOptimized (for example when you first get to the page, it should be optimized already).

frontend/src/buttons/OptimizeButton.css
```diff
@@ -0,0 +1,13 @@
+.optimized-btn-container {
```
mecolmg, line 1: I've mentioned this in @RCAVelez 's PR, but I'm not sure if it makes sense to have a buttons directory, when an application can have so many buttons. I'd probably put this in something like a `route` directory which contains components specific to rendering the route.

PR #35, opened 2020-06-25
Save btn

Cindyzhang977:
Create front end of button that would save the user's planned route. Onclick, button switches to a disabled button with text "Saved".

### DIFF
--- a/frontend/src/route/SaveButton.js
+++ b/frontend/src/route/SaveButton.js
@@ -1,0 +1,22 @@
+import React, { useState } from 'react';
+import Button from 'react-bootstrap/Button';
+
+/**
+ * Render button to save trip. If trip is saved, rendered disabled button with "Saved" text.
+ * @param {boolean|undefined} saved status of whether trip is saved.
+ */
+function SaveButton({saved}) {
+  const [isSaved, setIsSaved] = useState(saved);
+
+  if (isSaved) {
+    return (
+      <Button variant="primary" disabled>Saved</Button>
+    );
+  }
+
+  return (
+    <Button variant="primary" onClick={() => setIsSaved(true)}>Save Trip</Button>
+  );
+}
+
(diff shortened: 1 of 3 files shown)


frontend/src/route/SaveButton.js
```diff
@@ -0,0 +1,27 @@
+import React, { useState } from 'react';
```
zoebarth, line 1: Are you planning on putting all buttons in their own folder?

Cindyzhang977: No I think based on Colm's feedback, we'll just have buttons related to each feature inside of that feature's folder.

mecolmg: +1 to @Cindyzhang977 's response. Otherwise a folder called "buttons" or something similar just becomes a grab-bag of buttons, and it's not easy to find. The case where it would be fine to have a directory like that is if the button is truly generic and meant to be used in multiple places/features across the UI, in which case I would actually recommend a directory structure like `src/common/buttons/SaveButton.js`

frontend/src/route/SaveButton.js
```diff
@@ -0,0 +1,27 @@
...
+    );
+  }
+
+  handleSave() {
+    setIsSaved(true);
+  }
+
+  return (
+    <Button variant="primary" onClick={handleSave}>Save Trip</Button>
+  );
```
mecolmg, line 24: Consider:
```js
return (
  <Button variant="primary" onClick={handleSave} disabled={isSaved}>
    {isSaved ? 'Saved' : 'Save Trip'}
  </Button>
);
```
or
```js
const label = isSaved ? 'Saved' : 'Save Trip';
return (
  <Button variant="primary" onClick={handleSave} disabled={isSaved}>{label}</Button>
);
```

PR #38, opened 2020-06-26
add list of saved trips

zoebarth:
image

Create list of saved trips with name and edit and share button.

### DIFF
--- a/frontend/src/saved-trips/SavedTrip.js
+++ b/frontend/src/saved-trips/SavedTrip.js
@@ -3,20 +3,20 @@ import ListGroup from 'react-bootstrap/ListGroup';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faPencilAlt } from '@fortawesome/free-solid-svg-icons';
 import { faShare } from '@fortawesome/free-solid-svg-icons';
-import './SavedTrips.css';
+import styles from './SavedTrips.module.css';
 
 /** 
  * Creates single saved trip.
  */
 function SavedTrip({name}) {
   return (
-    <ListGroup.Item className="list-item">
+    <ListGroup.Item className={styles.listItem}>
       {name}
       <div>
-        <a href="/" title="Edit" className="icon">
+        <a href="/" title="Edit" className={styles.icon}>
           <FontAwesomeIcon icon={faPencilAlt} />
         </a>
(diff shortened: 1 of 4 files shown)


frontend/src/saved-trips/SavedTrip.js
```diff
@@ -0,0 +1,27 @@
...
+      {name}
+      <div>
+        <a href="/" title="Edit" className="icon">
+          <FontAwesomeIcon icon={faPencilAlt} />
+        </a>
+        <a href="/" title="Share" className="icon">
+          <FontAwesomeIcon icon={faShare}/>
+        </a>
+      </div>
+    </ListGroup.Item>
```
Cindyzhang977, line 23: Should we add an icon for just viewing the trip and not editing?

zoebarth: Good question! We can make the list item actionable so that if you just click the name you can view it

zoebarth: Let's discuss at standup tho

genedwards: +1 for making the name (ideally the whole bar to the left of the pencil icon) clickable

frontend/src/saved-trips/SavedTrips.js
```diff
@@ -0,0 +1,22 @@
...
+ * Creates list of saved trips.
+ */
+function SavedTrips({trips}) {
+  return (
+    <div className={styles.container}>
+      <h3>Saved Trips:</h3>
+      <ListGroup>
+        {trips.map(trip => (
+          <SavedTrip name={trip}/>
+        ))}
```
mecolmg, line 16: To make this cleaner, you could split it into a helper method, or store this in a variable.

frontend/src/saved-trips/SavedTrips.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import ListGroup from 'react-bootstrap/ListGroup';
+import styles from './SavedTrips.module.css';
+import SavedTrip from './SavedTrip.js';
+
+/** 
+ * Creates list of saved trips.
+ */
+function SavedTrips({trips}) {
```
mecolmg, line 9: Can you add a comment describing the type of trips? Assuming right now it's an array of strings.

PR #41, opened 2020-06-26
Route view

Cindyzhang977:
Compile components into route page view. Adjusted some styles on individual components. 
(Note: will modularize css in later PR)

image


frontend/src/RouteView.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
```
mecolmg, line 1: Have you all agreed that the root views will be stored in the src directory, or should it be put in a subdirectory (i.e. routeview or just route)?

zoebarth: I think we decided to store them directly in src

frontend/src/App.js
```diff
@@ -16,6 +18,7 @@ function App() {
         Click Me!
       </Button>
       <div>{text}</div>
+      <RouteView />
```
zoebarth, line 21: Are we planning on leaving this here?

Cindyzhang977: oh yeah we should prob remove the button

PR #45, opened 2020-06-29
Search View

zoebarth:
Add search view and add some functionality to navbar. Add css styling to move search bar down the page.
Screenshot 2020-06-30 at 10 20 18 AM - Display 2


frontend/src/SearchView.js
```diff
@@ -0,0 +1,26 @@
...
+ * @param {boolean} loggedIn - if the user is logged in.
+ */
+function SearchView({ loggedIn }) {
+  const trips = ['London', 'Paris'];
+
+  let savedTrips;
+  if (loggedIn) {
+    savedTrips = <SavedTrips trips={trips} />;
+  }
+
```
Cindyzhang977, line 17: It may be cleaner to write
```
const savedTrips = loggedIn ? <SavedTrips trips={trips} /> : <div></div>
```
Also maybe you can replace `<div></div>` with just `<></>` but I'm not sure.

frontend/src/SearchView.js
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    <div className={styles.container}>
+      <Search />
```
Cindyzhang977, line 20: To make the `Search` component the center of the screen, I think you can add `margin: auto` to its css.

frontend/src/SearchView.js
zoebarth: Yes, thank you!

frontend/src/SearchView.js
zoebarth: That didn't work for me :(

frontend/src/App.js
```diff
@@ -1,22 +1,20 @@
...
-  }
-
   return (
-    <div>
-      <Button onClick={handleClick} variant="secondary">
-        Click Me!
-      </Button>
-      <div>{text}</div>
-    </div>
+    <>
```
genedwards, line 8: wow, I didn't know this syntax for fragments

frontend/src/saved-trips/SavedTrips.js
```diff
@@ -8,25 +8,17 @@ import SavedTrip from './SavedTrip.js';
  * @param {string[]} trips - a list of the names of the trips
  */
 function SavedTrips({trips}) {
+  const tripList = trips.map(trip => 
+    <SavedTrip name={trip}/>
```
genedwards, line 12: does React complain about a missing key here? do you need a key? https://reactjs.org/docs/lists-and-keys.html#keys

frontend/src/SearchView.js
```diff
@@ -0,0 +1,25 @@
...
+import styles from './SearchView.module.css';
+
+/**
+ * Creates Search View component, with saved trips rendered if user is logged in.
+ * @param {boolean} loggedIn - if the user is logged in.
+ */
+function SearchView({ loggedIn }) {
+  const trips = ['London', 'Paris'];
+
+  const savedTrips = loggedIn ? <SavedTrips trips={trips} /> : <> </>;
```
genedwards, line 13: do you even need the fragment?
is the result the same as  `loggedIn ? <SavedTrips trips={trips} /> : null;`

frontend/src/navbar/Navbar.js
```diff
@@ -1,17 +1,24 @@
...
 import Nav from 'react-bootstrap/Nav';
 
 /** 
  * Creates Navbar component with login button.
  */
 function Navbar() {
+  const [text, setText] = useState('Sign In');
+
+  function handleLogin() {
+    const display = text === 'Sign In' ? 'Sign Out': 'Sign In';
```
genedwards, line 11: handling text like this scares me, no need to fix for the prototype, but please add a `TODO: Use enums instead of string.` so we can address this later

frontend/src/SearchView.js
zoebarth: Fixed it.

frontend/src/App.js
mecolmg: Yeah I just found this out the other day too! Much nicer than `<React.Fragment>`

frontend/src/App.js
```diff
@@ -1,25 +1,20 @@
-import React, { useState } from 'react';
-import Button from 'react-bootstrap/Button';
-
-import RouteView from './RouteView.js';
+import React from 'react';
+import { BrowserRouter as Router, Switch, Route } from 'react-router-dom';
+import SearchView from './SearchView.js';
+import Navbar from './navbar/Navbar.js';
 
 function App() {
```
mecolmg, line 6: I think you should move the login state to the App.js level, so you can pass the value to both the SearchView and Navbar components.

frontend/src/navbar/Navbar.js
```diff
@@ -1,17 +1,25 @@
-import React from 'react';
+import React, { useState } from 'react';
 import Nav from 'react-bootstrap/Nav';
 
 /** 
  * Creates Navbar component with login button.
  */
 function Navbar() {
+  const [text, setText] = useState('Sign In');
```
mecolmg, line 8: The state shouldn't be on this text, it should be on the boolean value of whether the user is logged in, which could just be a prop passed to this navbar. In order to update that state, you'd also have the Navbar take in an `onLoginChange` callback which you can use to update the state in App.js.

```js
// App.js
const [loggedIn, setLoggedIn] = useState(false);
return (
  <Navbar loggedIn={loggedIn} onLoginChange={setLoggedIn} />
  ...
);

// Navbar.js
function Navbar({loggedIn, onLoginChange}) {
  return (
    ...
    <Nav.Link onClick{() => onLoginChange(!loggedIn)} href="">{loggedIn ? 'Sign Out' : 'Sign In'}</Nav.Link>
    ...
   );
```

PR #47, opened 2020-06-29
Drag and Drop route items

Cindyzhang977:
Use [react-beautiful-dnd](https://github.com/atlassian/react-beautiful-dnd) to drag and drop cards in the route list to customize order. 

- fix height of route list, overflow is scrollable
- follow [this](https://egghead.io/lessons/react-reorder-a-list-with-react-beautiful-dnd) tutorial to implement drag/drop 
  - define entire `Route` component, which is the list of location cards as both the DragDropContext (See diagram at bottom for explanation of components) and Droppable
  - define each location card as Draggable
- follow [this](https://egghead.io/lessons/react-persist-list-reordering-with-react-beautiful-dnd-using-the-ondragend-callback) tutorial to persist the ordering of location cards after user has dragged a card 
  - in `handleOnDragEnd` in `Route.js`, create a duplicate of the original `placesList`, rearrange the ordering of places in that duplicate, then `setPlacesList` to that mutated duplicate
  - moved mapping of `placesList` to the return statement and made `placesList` a state so that a new list will be rendered every time the ordering changes
- minor CSS changes
  - on hover, overlay opacity increases
  - on hover, sort icon appears

On hover (mission peak):
image

Mid-drag:
image

reference react-dnd DragDropContext diagram:
image

### DIFF
--- a/frontend/src/route/LocationCard.js
+++ b/frontend/src/route/LocationCard.js
@@ -17,7 +17,7 @@ import { Draggable } from 'react-beautiful-dnd';
  */
 function LocationCard({location, description, image, index}) {
   // ref: https://egghead.io/lessons/react-reorder-a-list-with-react-beautiful-dnd
-  return(
+  return (
     <Draggable draggableId={`draggable-${index}`} index={index}>
       {(provided) => 
         <Container className={styles.locationCardContainer}
--- a/frontend/src/route/LocationCard.js
+++ b/frontend/src/route/LocationCard.js
@@ -5,17 +5,18 @@ import Row from 'react-bootstrap/Row';
 import Col from 'react-bootstrap/Col';
 import styles from './LocationCard.module.css';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import { faAngleUp, faAngleDown } from '@fortawesome/free-solid-svg-icons';
-
+import { faSort } from '@fortawesome/free-solid-svg-icons';
 import { Draggable } from 'react-beautiful-dnd';
 
 /**
  * Return a card component representing a place the user has selected. 
- * @param location the name of the place selected
- * @param description a description of @param location
- * @param image image of @param location
+ * @param {string} location the name of the place selected
+ * @param {string} description a description of @param location
+ * @param {string} image image source of @param location
+ * @param {number} index index of location in route list
  */
 function LocationCard({location, description, image, index}) {
+  // ref: https://egghead.io/lessons/react-reorder-a-list-with-react-beautiful-dnd
   return(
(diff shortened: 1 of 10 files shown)


zoebarth:
You may need to fix merge conflicts with master. If you can just leave the app.js the way it is in master, I will be able to add your page in the future.

frontend/src/route/LocationCard.js
```diff
@@ -5,40 +5,46 @@ import Row from 'react-bootstrap/Row';
...
- * @param image image of @param location
+ * @param {string} location the name of the place selected
+ * @param {string} description a description of @param location
+ * @param {string} image image source of @param location
+ * @param {number} index index of location in route list
  */
-function LocationCard({location, description, image}) {
+function LocationCard({location, description, image, index}) {
+  // ref: https://egghead.io/lessons/react-reorder-a-list-with-react-beautiful-dnd
   return(
```
mecolmg, line 20: nit: there should be a space after the return.

frontend/src/route/Route.js
```diff
@@ -1,27 +1,60 @@
...
-  const mockPlaces = places.map((place, index) => 
-    <Row key={index}>
-      <LocationCard 
-        location={place.location}
-        description={place.description}
-        image={place.image}
-      />
-    </Row>
-  );
+  const [placesList, setPlacesList] = useState(places);
```
mecolmg, line 11: nit: It seems like you added "List" to the end of this variable because "places" was already taken. If you want, you can rename the prop when destructing it like:
```js
function Route({places: initialPlaces}) {
  const [places, setPlaces] = useState(initialPlaces);
  ...
}
```

frontend/src/route/Route.js
```diff
@@ -1,27 +1,60 @@
...
+
+  // referece: https://egghead.io/lessons/react-persist-list-reordering-with-react-beautiful-dnd-using-the-ondragend-callback
+  function handleOnDragEnd(result) {
+    const {destination, source, draggableId } = result;
```
mecolmg, line 15: nit: You could destruct these in the function definition. 
```js
function handleOnDragEnd({destination, source, draggableId}) {...}
```

frontend/src/route/Route.js
```diff
@@ -1,27 +1,60 @@
...
+  }
 
+  // ref: https://egghead.io/lessons/react-reorder-a-list-with-react-beautiful-dnd
   return (
-    <Container>
-      {mockPlaces}
-    </Container>
+    <DragDropContext
+      onDragEnd={handleOnDragEnd}
+    >
```
mecolmg, line 37: nit: This can be in one line.

PR #59, opened 2020-06-30
Map pin add info window

Cindyzhang977:
Add info window to pop up when pins are clicked. Each window has the location name and image. It can also have description of the place if needed. 
image

(There was also a bracket issue in `package-lock.json` that was fixed.)


frontend/src/map/Map.js
```diff
@@ -31,11 +31,25 @@ function Map({ destinations, mode, centerLocation }) {
...
           const location = { lat: place.lat, lng: place.lng };
+          const infowindow = new googleMaps.InfoWindow({
+            content: `
+            <div>
+              <h4>${place.name}</h4>
+              <div>Short description of place if desired</div>
+              <div>
+                <img src="" alt="${place.name} Image" />
+              </div>
+            </div>
```
mecolmg, line 42: nit: one more tab.

frontend/src/map/Map.js
```diff
@@ -31,11 +31,25 @@ function Map({ destinations, mode, centerLocation }) {
...
+            `
+          });
           const marker = new googleMaps.Marker({
             position: location,
             map,
             title: place.name,
           });
+          marker.addListener('click', function() {
```
mecolmg, line 50: nit: prefer arrow functions for callbacks. 
https://google.github.io/styleguide/jsguide.html#features-functions-arrow-functions

PR #61, opened 2020-07-01
url paths display ExploreView, and RouteView

RCAVelez:
** What's new:**
/explore shows ExploreView
/route shows RouteView


frontend/src/ExploreView.js
```diff
@@ -13,7 +13,7 @@ function Explore({ images }) {
     <div className={styles.exploreContainer}>
       <div className={styles.attractionsSection}>
         <AttractionsPanel attractionUrls={images} />
-        <NavButton link={'/route'} label={'Show Route'} />
+        <NavButton href={'/route'} label={'Show Route'} />
       </div>
       <div className={styles.mapContainer}>
         <Map destinations={''} mode={'pins'} />
```
Cindyzhang977, line 19: I don't think you need `{ }` when it's a string. Just `href="/route"` , `mode="pins"`, etc. should be okay.

PR #93, opened 2020-07-06
add navigation link to saved trip

zoebarth:
When user clicks on saved trip, navigate to the route view of the trip.

For MVP purposes, only use index of the trip until database is created.

### DIFF
--- a/frontend/src/saved-trips/SavedTrips.js
+++ b/frontend/src/saved-trips/SavedTrips.js
@@ -8,7 +8,9 @@ import SavedTrip from './SavedTrip.js';
  * @param {string[]} trips - a list of the names of the trips
  */
 function SavedTrips({ trips }) {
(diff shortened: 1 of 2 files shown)


frontend/src/saved-trips/SavedTrips.js
```diff
@@ -8,7 +8,9 @@ import SavedTrip from './SavedTrip.js';
  * @param {string[]} trips - a list of the names of the trips
  */
 function SavedTrips({ trips }) {
-  const tripList = trips.map((trip, index) => <SavedTrip key={index} name={trip} />);
+  const tripList = trips.map((trip, index) => (
+    <SavedTrip key={index} id={index} name={trip} />
+  ));
```
Cindyzhang977, line 13: nit: are the parenthesis around the SavedTrip component needed?

PR #96, opened 2020-07-07
Github CI compiles backend Maven build.

genedwards:
Run `mvn package` on push to github. Goal is to verify that the Java code compiles.

Tested with commit 6552066: https://github.com/googleinterns/step88-2020/runs/846066529
image

### DIFF
--- a/frontend/src/RouteView.js
+++ b/frontend/src/RouteView.js
@@ -19,6 +19,8 @@ function RouteView() {
   const [isOptimized, setIsOptimized] = useState(false);
   const [isSaved, setIsSaved] = useState(false);
   const [places, setPlaces] = useState(MOCK_DATA);
+  // TODO: Remove temporarily disabled linter.
+  // eslint-disable-next-line no-unused-vars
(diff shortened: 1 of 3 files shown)


frontend/src/RouteView.js
```diff
@@ -19,6 +19,8 @@ function RouteView() {
   const [isOptimized, setIsOptimized] = useState(false);
   const [isSaved, setIsSaved] = useState(false);
   const [places, setPlaces] = useState(MOCK_DATA);
+  // TODO: Remove temporarily disabled linter.
+  // eslint-disable-next-line no-unused-vars
```
genedwards, line 23: I'm not sure how this got into master branch

frontend/src/RouteView.js
```diff
@@ -19,6 +19,8 @@ function RouteView() {
...
   const [optimizedOrder, setOptimizedOrder] = useState(null);
```
genedwards, line 24: I'm not sure how this change got into master branch when the linter fails

PR #97, opened 2020-07-07
Frontend style

zoebarth:
Change styling of search page as shown below
image


frontend/src/search/Search.module.css
```diff
@@ -1,13 +1,44 @@
...
+
+.grey {
+  color: #6c757d;
+}
+
+.red {
+  color: #ea4335;
+}
+
+.searchBar:global(.form-control) {
```
genedwards, line 22: what does this do? what's this for?

zoebarth: Colm suggested using this instead of the !important tag.

zoebarth: https://github.com/css-modules/css-modules#exceptions

frontend/src/search/Search.js
```diff
@@ -17,7 +17,14 @@ function Search() {
   return (
     <div className={styles.searchContainer}>
       <div className={styles.whereTo}>
-        <h1>Where to?</h1>
+        <h1 className={styles.text}>
+          <span className={styles.blue}>W</span>
```
mecolmg, line 21: I was very disappointed to find out there's no such css selector for `:nth-letter` which would've made this soooooo much nicer.

frontend/src/search/Search.module.css
mecolmg: @genedwards in gss this is the equivalent to:
```
.SearchBar.form-control
```
It's to ensure increased specificity over the overriden bootstrap component styles.

frontend/src/search/Search.module.css
```diff
@@ -1,13 +1,44 @@
...
+  color: #34a853;
+}
+
+.grey {
+  color: #6c757d;
+}
+
+.red {
+  color: #ea4335;
+}
```
mecolmg, line 20: nit: You don't have to sort the class names in your css file. It's preferable to keep related things together like these colors styles.

frontend/src/search/Search.js
mecolmg: Also, this is one place I would've definitely considered inline styles.

frontend/src/search/Search.module.css
genedwards: ok, so I take this as `:local` to be obfuscated, and `:global` as not obfuscated.
So, `.searchBar:global(.form-control)` would be something like `.searchBar-123abc.form-control`, and we do this because the FormControl itself has gss's equivalent of `.exportClass`. Ok.
Thanks for the explanation

PR #105, opened 2020-07-08
Mst Alg

Cindyzhang977:
Implement Prim's algorithm to construct a minimum spanning tree of a given graph.

- implement `getMst` method in `TspOptimizer` which runs Prim's algorithm using the given `source` as the starting vertex
- add `equals` and `getEndpoints` methods in `Edge`
- add more junit tests for `getMst`
  - create mst for `K4` graph 
  - create helper function `checkGraphEquals`
    - iterate through the vertices and check if the edges each vertex maps to are the same as the expected mst
    - sort each edge so the ArrayLists are equal if the contents are the same

### DIFF
--- a/backend/src/test/java/com/google/sps/TspOptimizerTest.java
+++ b/backend/src/test/java/com/google/sps/TspOptimizerTest.java
@@ -89,7 +89,7 @@ public final class TspOptimizerTest {
   @Test
   @Ignore
   public void optimize_TRIANGLE_pickRandomSourceAttraction() {
-    // Use triangle graph. Choose a random Attraction as source. 
+    // Use triangle graph. Choose a random Attraction as source.
     // Algorithm should return a path with the two shortest edges A--B--C.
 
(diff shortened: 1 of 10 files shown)


zoebarth:
Make sure you run prettier one more time

backend/src/test/java/com/google/sps/TspOptimizerTest.java
```diff
@@ -36,21 +37,42 @@
...
   private static final HashMap<Attraction, ArrayList<Edge>> K4 = create_K4();
+  private static final HashMap<Attraction, ArrayList<Edge>> K4_MST = create_K4_MST();
 
   @Test
-  @Ignore
-  public void mst_TRIANGLE() {
-    // Generate the MST of the TRIANGLE graph
+  public void mst_TRIANGLE_A() {
+    // Generate the MST of the TRIANGLE graph starting at vertex A
+
```
genedwards, line 49: nit: remove blank line here, and the other 3 tests below

PR #111, opened 2020-07-08
Dfs Alg

Cindyzhang977:
Implement `dfs` function, which returns an arraylist of `Attraction`s in the order of the dfs traversal of the graph's mst. 

- Include helper function `dfsHelper`
  - recursively called to visit all vertices depth first
  - maintains `visited` set to keep track of all visited vertices 
- add `dfs` tests on `TRIANGLE` and `K4` graphs

### DIFF
--- a/backend/src/main/java/com/google/sps/TspOptimizer.java
+++ b/backend/src/main/java/com/google/sps/TspOptimizer.java
@@ -95,8 +95,32 @@ public final class TspOptimizer {
     return mst;
   }
 
-  // Visible for testing
+  /**
+   * Return list of attractions in pre-order DFS traversal order.
+   * @param source the vertex on which to start DFS
+   * @param graph the MST on which to run DFS
+   */
   static ArrayList<Attraction> dfs(Attraction source, HashMap<Attraction, ArrayList<Edge>> graph) {
-    return null;
+    return dfsHelper(source, graph, new HashSet<>());
+  }
+
+  /**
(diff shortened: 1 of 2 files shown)


backend/src/main/java/com/google/sps/TspOptimizer.java
```diff
@@ -95,8 +95,32 @@
...
+  }
+
+  /**
+   * Helper function for dfs.
+   * @param curr the vertex that is currently being visited in the DFS traversal
+   * @param graph the MST on which to run DFS
+   * @param visited set of visited vertices
+   */
+  private static ArrayList<Attraction> dfsHelper(
+      Attraction curr, HashMap<Attraction, ArrayList<Edge>> graph, HashSet<Attraction> visited) {
```
genedwards, line 114: Why are we passing `visited` here, what's the use case?

Cindyzhang977: This is so dfs only keeps visiting attractions that have not yet been traversed. Otherwise, I think since the graph has lots of cycles, dfs would never end.

genedwards: dfs() passes a new HashSet, I was wondering why can't dfsHelper just create a new HashSet  instead? unless there's a use case where you plan to call dfsHelper and pass in `visited` that's not a new HashSet

PR #112, opened 2020-07-09
Tsp optimization alg

Cindyzhang977:
Implement `optimize` functionality of TSP approximation. Add/edit Junit tests for `optimize` functions.

- `optimize(source, graph)` runs the TSP algorithm from a specified source
- `optimize(graph)` uses the first vertex in the graph HashMap keySet as `source` and then calls `optimize(source, graph)`
- implement helper function `getOptimalOrdering`
  - removes the heaviest edge from the dfs ordering to produce an optimized ordering

Pseudocode for optimize (with source):
```
mst <-- getMst(source, graph) 
dfsOrder <-- dfs pre-order of mst starting at source
find where the heaviest edge is in the dfsOrder cycle, 
    where the last attraction has an edge to first attraction (i.e: A-B-C-D-A)
startIndex <-- index of attraction to start trip route (index of attraction after heaviest edge)
return ArrayList<Attraction> of (startIndex)--(lastAttraction)--(firstAttraction)--(startIndex - 1)
``` 

Before: 
image

After: 
image

### DIFF
--- a/backend/src/main/java/com/google/sps/TspOptimizer.java
+++ b/backend/src/main/java/com/google/sps/TspOptimizer.java
@@ -21,7 +21,8 @@ import java.util.PriorityQueue;
 
 public final class TspOptimizer {
   /**
-   * Call optimize(source, graph) after determining a source Attraction from which to run MST algorithm.
+   * Call optimize(source, graph) after determining a source Attraction from which to run MST
+   * algorithm.
    * @return list of attractions in optimal visiting order
    */
   public static ArrayList<Attraction> optimize(HashMap<Attraction, ArrayList<Edge>> graph) {
@@ -47,7 +48,8 @@ public final class TspOptimizer {
    * @param graph the mst that contains the attractions and that was dfs traversed
    * @return arraylist containing the attractions in optimized order
    */
-  private static ArrayList<Attraction> getOptimalOrdering(ArrayList<Attraction> attractions, HashMap<Attraction, ArrayList<Edge>> graph) {
+  private static ArrayList<Attraction> getOptimalOrdering(
+      ArrayList<Attraction> attractions, HashMap<Attraction, ArrayList<Edge>> graph) {
     // determine heaviest edge
     long maxDistance = 0;
     int startIndex = 0;
@@ -64,7 +66,8 @@ public final class TspOptimizer {
     }
 
     // create optimized ordering
-    ArrayList<Attraction> optimizedOrder = new ArrayList<>(attractions.subList(startIndex, attractions.size()));
+    ArrayList<Attraction> optimizedOrder =
+        new ArrayList<>(attractions.subList(startIndex, attractions.size()));
     optimizedOrder.addAll(attractions.subList(0, startIndex));
     return optimizedOrder;
   }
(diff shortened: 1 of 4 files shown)


backend/src/main/java/com/google/sps/TspOptimizer.java
```diff
@@ -21,15 +21,13 @@
 
 public final class TspOptimizer {
   /**
-   * Call optimize(source) after determining a source Attraction from which to run MST algorithm.
+   * Call optimize(source, graph) after determining a source Attraction from which to run MST
+   * algorithm.
    * @return list of attractions in optimal visiting order
    */
   public static ArrayList<Attraction> optimize(HashMap<Attraction, ArrayList<Edge>> graph) {
```
zoebarth, line 28: Add @param tags for all parameters

backend/src/main/java/com/google/sps/TspOptimizer.java
```diff
@@ -39,7 +37,39 @@
    */
```
zoebarth, line 37: Add @param tag for graph

backend/src/main/java/com/google/sps/TspOptimizer.java
```diff
@@ -95,8 +125,32 @@
     return mst;
   }
 
-  // Visible for testing
+  /**
+   * Return list of attractions in pre-order DFS traversal order.
+   * @param source the vertex on which to start DFS
+   * @param graph the MST on which to run DFS
+   */
```
zoebarth, line 132: add @return tag

backend/src/main/java/com/google/sps/TspOptimizer.java
```diff
@@ -95,8 +125,32 @@
...
-    return null;
+    return dfsHelper(source, graph, new HashSet<>());
+  }
+
+  /**
+   * Helper function for dfs.
+   * @param curr the vertex that is currently being visited in the DFS traversal
+   * @param graph the MST on which to run DFS
+   * @param visited set of visited vertices
+   */
```
zoebarth, line 142: missing a @return tag

backend/src/main/java/com/google/sps/TspOptimizer.java
```diff
@@ -21,25 +21,57 @@
...
+        if (e.getOtherEndpoint(curr).equals(next)) {
+          maxDistance = Math.max(e.getDistance(), maxDistance);
+          startIndex = e.getDistance() == maxDistance ? nextIndex : startIndex;
+        }
+      }
+    }
+
+    // create optimized ordering
+    ArrayList<Attraction> optimizedOrder =
+        new ArrayList<>(attractions.subList(startIndex, attractions.size()));
```
mecolmg, line 72: nit: It's generally better if you're creating a library to just store and return a interface like `List` instead of `ArrayList` so that you're not forcing yourself and clients to use an `ArrayList` when any implementer of `List` would do. Overall, any clients of your library shouldn't care about what specific implementation of `List` is being used behind the scenes.

Then this could be
```java
List<Attraction> optimizedOrder = attractions.subList(startIndex, attractions.size());
```

https://stackoverflow.com/questions/11683044/return-arraylist-or-list

You're in pretty deep now, but it would be a nice little refactor.

PR #113, opened 2020-07-09
Redirect authentication

zoebarth:
Add redirect url to authentication so that when user logs in on a page they are redirected to the same page upon successful login or logout. 

Add index key to attractions panel to remove warning.

### DIFF
--- a/frontend/src/App.js
+++ b/frontend/src/App.js
@@ -1,5 +1,5 @@
 import React, { useState, useEffect } from 'react';
-import { BrowserRouter as Router, Switch, Route } from 'react-router-dom';
+import { BrowserRouter as Router, Switch, Route, useLocation } from 'react-router-dom';
 import SearchView from './SearchView.js';
 import ExploreView from './ExploreView.js';
 import RouteView from './RouteView.js';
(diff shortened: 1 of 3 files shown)


frontend/src/attractions-panel/AttractionsPanel.js
```diff
@@ -10,8 +10,8 @@ import styles from './AttractionsPanel.module.css';
 function AttractionsPanel({ attractionUrls }) {
   return (
     <div className={styles.attractionsPanel}>
-      {attractionUrls.map((url) => (
-        <Attraction imageUrl={url} key={url} />
+      {attractionUrls.map((url, index) => (
+        <Attraction imageUrl={url} key={index} />
```
genedwards, line 14: 👍  thanks for fixing, could've been a separate PR though

frontend/src/App.js
```diff
@@ -10,30 +10,42 @@ import testImg from './attraction/Attraction.png';
...
+
+  useEffect(
+    () => {
+      fetch(`/api/v1/auth?redirect=${redirectUrl}`)
+        .then((response) => response.json())
+        .then(({ loggedIn, loginUrl, logoutUrl }) =>
+          onChange({ ready: true, loggedIn, loginUrl, logoutUrl })
+        );
+    },
+    /* Don't refetch on rerender. */ [redirectUrl, onChange]
```
genedwards, line 45: So I assume `redirectUrl` is here because if the url changes, we fetch again with the new redirectUrl.
But why is `onChange` here?

zoebarth: I put it there to get the linter to stop giving me an error. Please let me know if there is a better way to handle it.

mecolmg: Did it not just work with an empty array? This doesn't make sense to me.

mecolmg: Sorry for leading you down the rabbit hole here. This appears to be the correct way to do this: https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies

PR #116, opened 2020-07-10
Move save button

zoebarth:
Move save trip button as according to the mockup. Button is positioned absolutely over the navbar but is still part of the RouteView component in order to preserve state. Button is only visible when user is logged in.

If the user is logged in:
image

If the user is not logged in:
image

### DIFF
--- a/frontend/src/RouteView.js
+++ b/frontend/src/RouteView.js
@@ -15,7 +15,7 @@ import { MOCK_DATA } from './route/mockData.js';
 /**
  * Render the route page with list of locations in order and directions on a map between the locations.
  */
-function RouteView() {
+function RouteView({ loggedIn }) {
   const [isOptimized, setIsOptimized] = useState(false);
   const [isSaved, setIsSaved] = useState(false);
   const [places, setPlaces] = useState(MOCK_DATA);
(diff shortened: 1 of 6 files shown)


frontend/src/navbar/Navbar.js
```diff
@@ -25,7 +25,12 @@ function Navbar({ authState }) {
           href={authState.loggedIn ? authState.logoutUrl : authState.loginUrl}
           className={styles.rightLink}
         >
-          {authState.loggedIn ? 'Sign Out' : 'Sign In'}
+          <Switch>
+            <Route path="/route">
+              {authState.loggedIn ? '' : 'Sign in to save trips'}
+            </Route>
```
Cindyzhang977, line 31: There might be merge conflicts here, since this was deleted in some later PRs.

frontend/src/route/SaveButton.module.css
```diff
@@ -0,0 +1,12 @@
...
+  right: 16px;
+  top: 8px;
+}
+
+.loggedOut {
+  position: absolute;
+  right: 16px;
+  top: 8px;
+  visibility: hidden;
+}
```
Cindyzhang977, line 12: Instead of `loggedIn`/`loggedOut`, since most of it is the same, can we have one class for position properties and a class that is added to toggle visibility?

genedwards: I would go a step further and just not render it:
```
{isLoggedIn &&
  <Button ...>
    {isSaved ? 'Saved' : 'Save Trip'}
  </Button>
}
```

frontend/src/navbar/Navbar.js
zoebarth: Okay, we'll figure it out.

frontend/src/route/SaveButton.module.css
zoebarth: Done.

frontend/src/RouteView.js
```diff
@@ -50,33 +50,33 @@ function RouteView() {
...
-            <Map mode="directions" places={places} centerLocation={places[0]} />
-          </Row>
-          <Row>
-            <SaveButton isSaved={isSaved} save={save} />
-          </Row>
-        </Col>
-      </Row>
-    </Container>
+    <>
+      <SaveButton isSaved={isSaved} save={save} isLoggedIn={loggedIn} />
```
mecolmg, line 54: Instead of rendering this in the RouteView and positioning it absolutely, have you considered rendering it in the Nav bar based on the route?

zoebarth: I considered that and actually went down that rabbit hole for a good hour but then realized how difficult it is to pass all of the state.

mecolmg: Spoke offline. The conflict would be that a lot of state would need to be lifted up to App.js in order to do this. Another option would be to render the NavBar in each of the views instead of having one for all of the views, so that the buttons and links in them can be conditioned without having to use a router Switch, but that shouldn't be a blocker for this PR.

PR #123, opened 2020-07-13
Routing edge cases

Cindyzhang977:
Address routing edge cases so app doesn't error. 

- disable "Show Route" button until at least one attraction is selected 
- when an attraction doesn't have an image, skip it and move on to the next attraction 
- increase nearby search radius to 10000 meters (~ 6 miles)
- display text "No Images Found" if no attractions with images are found
- min-width and height of images panel so it doesn't collapse when images are loading


frontend/src/ExploreView.js
```diff
@@ -70,21 +70,30 @@ function Explore() {
...
-                  attraction.selected ? styles.selectedImage : ''
-                }`}
-                src={attraction.photoUrl}
-                alt=""
-              />
-            </div>
-          ))}
-        </div>
-        <Button className={styles.routeButton} onClick={() => handleRouting(history)}>
+        {allAttractions.length === 0 && calledAPI ? (
```
zoebarth, line 73: Have you considered making this a function or constant for readability?

frontend/src/ExploreView.js
```diff
@@ -150,6 +159,9 @@ function Explore() {
   function getAllAttractions(attractions) {
     const newAllAttractions = [];
     for (const attraction of attractions) {
+      if (!Object.prototype.hasOwnProperty.call(attraction, 'photos')) {
+        continue;
+      }
       if (Object.prototype.hasOwnProperty.call(attraction.photos[0], 'getUrl')) {
```
mecolmg, line 165: What's with the `Object.prototype.hasOwnPropoerty.call`'s? You can either do `'photos' in attraction`,  `attractiono.photos !== undefined`, or `attraction.hasOwnProperty('photos')`

frontend/src/ExploreView.js
```diff
@@ -8,6 +8,7 @@ import { useLocation, useHistory } from 'react-router-dom';
  * Explore view with selectable attraction images and map
  */
 function Explore() {
+  const [calledAPI, setCalledAPI] = useState(false);
```
mecolmg, line 11: nit: generally the state you care about is whether it's loading or not, so consider just calling this loading. Otherwise it's not clear what you'll use this for.

frontend/src/ExploreView.js
```diff
@@ -150,6 +159,9 @@ function Explore() {
   function getAllAttractions(attractions) {
     const newAllAttractions = [];
     for (const attraction of attractions) {
+      if (!Object.prototype.hasOwnProperty.call(attraction, 'photos')) {
```
genedwards, line 162: I would use `if (!attraction.photos)` or one of the options Colm suggested below.

PR #126, opened 2020-07-13
Share btn & modal

Cindyzhang977:
Implement a modal that shows when the "Share" button is clicked.

- change `SaveButton.js` to be `SaveShareButtons.js` to include both buttons 
- move `loggedIn` logic to `RouteView.js` in order to render both buttons 
- implement ability to copy the url in the input field on the modal with the click of a button 
  - url is the full current url loaded in the browser 
- add padding to navbar


frontend/src/RouteView.js
```diff
@@ -49,9 +58,49 @@ function RouteView({ loggedIn }) {
     setIsSaved(false);
   }
 
+  function copyToClipboard(e) {
```
genedwards, line 61: 👍 very nice

frontend/src/RouteView.js
```diff
@@ -49,9 +58,49 @@ function RouteView({ loggedIn }) {
...
+              </Button>
+            </Row>
+          </Container>
+        </Modal.Body>
+      </Modal>
+      {loggedIn ? (
+        <SaveShareButtons isSaved={isSaved} save={save} share={handleShareShow} />
+      ) : (
+        <></>
+      )}
```
genedwards, line 103: Maybe just `{ loggedIn && (<SaveShareButtons .../>) }` instead of the empty fragment.

PR #128, opened 2020-07-14
Add checkmark, Explore Page css

Cindyzhang977:
Add checkmarks to images when selected and center Explore page.

- refactor each image into a bootstrap `Card`
- add a Card overlay that increases in opacity on hover and when image is selected 
- center `ExploreView` with bootstap `Container`, `Row`, `Col`
  - remove `mapContainer` styles
  - only looks nice when viewed on monitor, not responsive so far 

image

### DIFF
--- a/frontend/src/ExploreView.module.css
+++ b/frontend/src/ExploreView.module.css
@@ -3,7 +3,11 @@
 }
 
 .attractionContainer {
-  position: relative;
+  margin: 10px;
+}
+
+.attractionContainer:hover {
+  cursor: pointer;
 }
 
 .attractionsSection {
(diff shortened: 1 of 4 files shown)


frontend/src/ExploreView.module.css
```diff
@@ -18,22 +20,15 @@
 
 .attraction {
   height: 300px;
-  margin: 10px;
+  object-fit: cover;
```
genedwards, line 23: thank you! while not supported in IE, this looks much nicer for all other browsers

frontend/src/ExploreView.module.css
```diff
@@ -18,22 +20,15 @@
...
-.selectedImage {
-  border: solid green;
-  border-width: 10px;
-}
-
-.mapContainer {
-  margin-right: 100px;
-  margin-top: 30px;
+.check {
+  color: whitesmoke;
```
zoebarth, line 28: Whitesmoke >>>>>>> white

PR #129, opened 2020-07-14
map info window styles

Cindyzhang977:
Fix css in map info windows that pop up when the pin is clicked. I had to use inline styles for the css to show up on the map. 

image

### DIFF
--- a/frontend/src/map/Map.js
+++ b/frontend/src/map/Map.js
@@ -18,10 +18,10 @@ function Map({ attractions, mode, centerLocation, google, onReady, view }) {
       const infowindow = new google.maps.InfoWindow({
         content: `
           <div>
-            <h4 style="color:#383838;">${attraction.name}</h4>
-            <div style="margin-bottom:10px;">Short description of attraction if desired</div>
+            <h4 class=${styles.infowindowName}>${attraction.name}</h4>
+            <div class=${styles.infowindowDescription}>Short description of attraction if desired</div>
             <div>
-              <img style="height:200px;" src="${attraction.photoUrl}" alt="${attraction.name} Image" />
+              <img class=${styles.infowindowImg} src="${attraction.photoUrl}" alt="${attraction.name} Image" />
             </div>
           </div>
         `,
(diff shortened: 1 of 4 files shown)


frontend/src/map/Map.js
```diff
@@ -18,10 +18,10 @@ function Map({ attractions, mode, centerLocation, google, onReady, view }) {
       const infowindow = new google.maps.InfoWindow({
         content: `
           <div>
-            <h4>${attraction.name}</h4>
-            <div>Short description of attraction if desired</div>
+            <h4 style="color:#383838;">${attraction.name}</h4>
```
genedwards, line 21: I feel we should be able to pass a class name
```JS
<h4 class="${styles.map-popup-title}">${attraction.name}</h4>
```

zoebarth, line 21: Why did you have to do inline styling?

Cindyzhang977: I did that, but then the styles don't show up. The className shows in the html of the inspector but not the css. This might have something to do with google maps styles.

Cindyzhang977: The styles don't show up otherwise. It's very strange.

genedwards: This is real HTML and not JSX, can you try `class` like in my example instead of `className`?

Cindyzhang977: Ah ok, it has been fixed

frontend/src/map/Map.js
```diff
@@ -18,10 +18,10 @@ function Map({ attractions, mode, centerLocation, google, onReady, view }) {
       const infowindow = new google.maps.InfoWindow({
         content: `
           <div>
-            <h4>${attraction.name}</h4>
-            <div>Short description of attraction if desired</div>
+            <h4 class=${styles.infowindowName}>${attraction.name}</h4>
```
mecolmg, line 21: Do these not need quotes around them? I'm really surprised if you don't!

frontend/src/map/Map.module.css
```diff
@@ -5,3 +5,15 @@
   max-height: 900px;
   max-width: 700px;
 }
+
+.infowindowName {
```
mecolmg, line 9: nit: can we follow camelCase and make this infoWindowName (same below)?

frontend/src/map/Map.js
Cindyzhang977: It does :)

PR #130, opened 2020-07-14
Navbar css

zoebarth:
Move edit button up into navbar, and fix bug that causes clicking anywhere on the navbar to bring you back.

image
image
image

### DIFF
--- a/frontend/src/navbar/Navbar.module.css
+++ b/frontend/src/navbar/Navbar.module.css
@@ -1,6 +1,6 @@
 .navbar {
   display: flex;
   justify-content: space-between;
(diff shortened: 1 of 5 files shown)


frontend/src/navbar/Navbar.module.css
```diff
@@ -1,13 +1,6 @@
 .navbar {
   display: flex;
-  padding: 12px 32px;
+  justify-content: space-between;
+  padding: 12px 0px;
```
Cindyzhang977, line 4: very very nit: don't need `px` for `0` (can just be `padding: 12px 0;`)

PR #135, opened 2020-07-14
style route page

zoebarth:
Make height of map and route container 70vh so it fits on the screen. Adjust padding and margins as necessary; remove hard coded placeholder text.

image

### DIFF
--- a/frontend/src/route/LocationCard.module.css
+++ b/frontend/src/route/LocationCard.module.css
@@ -1,6 +1,6 @@
 .locationCardContainer {
   color: white;
-  margin: 16px;
+  margin: 0 16px 16px;
   max-width: 600px;
 }
 
--- a/frontend/src/route/LocationCard.module.css
+++ b/frontend/src/route/LocationCard.module.css
@@ -1,6 +1,6 @@
 .locationCardContainer {
   color: white;
-  margin: 0 16px 16px;
+  margin-bottom: 16px;
   max-width: 600px;
 }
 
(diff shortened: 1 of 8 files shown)


frontend/src/route/LocationCard.module.css
```diff
@@ -1,6 +1,6 @@
 .locationCardContainer {
   color: white;
-  margin: 16px;
+  margin: 0 16px 16px;
```
genedwards, line 3: This container has a vertical scrollbar because the contents are too wide.
Can you change this to `margin: 16px 0` instead?
image

frontend/src/trip-name/TripName.js
```diff
@@ -13,7 +13,7 @@ import styles from './TripName.module.css';
  */
 function TripName({ tripName }) {
   const [isEditing, setIsEditing] = useState(false);
-  const [name, setName] = useState(tripName ? tripName : 'Trip Name');
+  const [name, setName] = useState(tripName ? tripName : '');
```
genedwards, line 16: you can do `useState(tripName || '')`

frontend/src/trip-name/TripName.module.css
```diff
@@ -31,7 +31,7 @@
   color: #8f8f8f;
   display: none;
   font-size: 0.9rem;
-  margin: auto;
+  margin: -6px auto 0px auto;
```
genedwards, line 34: I don't really like negative margin, it's a sign that something else is wrong.
I got this to work, please try it out:
```CSS
.tripNameContainer :global(.form-group) {
  margin-bottom: 0;
}

.saveBtn {
  ...
  margin: 12px auto;
  ...
}

// also, delete styling for .saveRow
```

frontend/src/route/LocationCard.module.css
RCAVelez: ^

PR #136, opened 2020-07-14
fix attractions bug

Cindyzhang977:
Change css of `Card` to match the 300px height of attraction images, and increase search radius to 10000 meters. There is a bug that sometimes only two attraction images would load for a search like "Paris" or "London". I was able to fix the css, but not the search results.

bug: 
image

fix: 
image

### DIFF
--- a/frontend/src/ExploreView.js
+++ b/frontend/src/ExploreView.js
@@ -52,7 +52,7 @@ function Explore() {
         placesService.nearbySearch(
           {
             location: coordinates,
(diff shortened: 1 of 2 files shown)


frontend/src/ExploreView.js
```diff
@@ -80,6 +80,7 @@ function Explore() {
     );
   };
 
+  console.log(initialAttractions);
```
genedwards, line 83: please remove

PR #150, opened 2020-07-15
Style explore page

zoebarth:
image
Add extra styling to explore page.


frontend/src/ExploreView.module.css
```diff
@@ -21,15 +21,18 @@
...
-  height: 15vw;
+  height: 13vw;
   object-fit: cover;
-  width: 15vw;
+  width: 13vw;
+}
+.centerCheck {
+  justify-content: center;
+  display: flex;
+  align-items: center;
```
Cindyzhang977, line 32: alphabetize

zoebarth: Done

zoebarth:
> Interestingly, you see less photos on larger screens. @2350px wide, it shows only 2 columns vs 3 columns for smaller screens. If you don't have a screen that large, you can simulate it by zooming out with Ctrl + `-`.
> It's an improvement overall for most other cases. LGTM

I removed a max width to fix that.

PR #152, opened 2020-07-15
push to history route changes, move handleRouting function

Cindyzhang977:
- push to history the new route order every time it's manually updated
  - sharing link now shows the same route order
- move `setAttractions` to `onManualPlaceChange` function in `RouteView` to avoid passing another prop to `Route`
- move `handleRouting` function in `ExploreView` and `RouteView` to `routingUtils.js` (renamed from `parameterUtils.js`)
  - add parameter `page` to indicate whether to route to `/explore` or `/route`
  - add parameter `attractions` for selected attractions, `tripObject` for the current trip object

### DIFF
--- a/frontend/src/RouteView.js
+++ b/frontend/src/RouteView.js
@@ -8,13 +8,14 @@ import Modal from 'react-bootstrap/Modal';
 import styles from './RouteView.module.css';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faClone } from '@fortawesome/free-solid-svg-icons';
+import { getQueryParameters, handleRouting } from './routingUtils.js';
 
 import Map from './map/Map.js';
 import Route from './route/Route.js';
 import OptimizeButton from './route/OptimizeButton.js';
 import SaveShareButtons from './route/SaveShareButtons.js';
 import TripName from './trip-name/TripName.js';
-import { getQueryParameters } from './parameterUtils.js';
+
 /**
  * Render the route page with list of locations in order and directions on a map between the locations.
  */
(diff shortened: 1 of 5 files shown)


frontend/src/RouteView.js
```diff
@@ -57,26 +58,20 @@ function RouteView({ loggedIn }) {
...
   }
 
-  function onManualPlaceChange() {
+  function onManualPlaceChange(newAttractions) {
+    setAttractions(newAttractions);
     setIsOptimized(false);
     setIsSaved(false);
+    tripObject.selectedAttractions = newAttractions;
+    const url = '?trip=' + encodeURIComponent(JSON.stringify(tripObject));
+    history.push(`/route${url}`);
```
genedwards, line 67: can we use handleRouting here?

frontend/src/routingUtils.js
```diff
@@ -0,0 +1,22 @@
...
+ * Creates url and navigates to /{page}?trip=
+ * @param {object} history used to route dom with react
+ * @param {string} page the page to which to route (explore|route)
+ * @param {object} tripObject json object that contains trip information
+ * @param {object} attractions list of selected attractions
+ */
+export const handleRouting = (history, page, tripObject, attractions) => {
+  tripObject.selectedAttractions = attractions;
+  const url = '?trip=' + encodeURIComponent(JSON.stringify(tripObject));
+  history.push(`/${page}${url}`);
```
genedwards, line 21: FYI, one thing to think about: each time you rearrange the order, there's a new history entry, if you rearranged 10 times, you'll need to hit back 10 times to get back to the previous page.
An alternate approach would be to replace history when you rearrange, so that the "previous" page is what it initially loaded with. I'm not saying that's better though, it's just a different approach.
In our case we have the "Edit Attractions" link at the top of the page for users who really want to get back to the explore page so maybe pushing to history each time we rearrange makes sense for us.

PR #153, opened 2020-07-15
search css

Cindyzhang977:
- search icon changed to fontawesome icon positioned absolutely on top of search bar 
- add `max-width` to the saved trips 
- add colors as constants in `App.css` so Google colors and gray can be used consistently across the app
  - ex: used `color: var(--blue);` to set to color to Google blue 

image

### DIFF
--- a/frontend/src/App.css
+++ b/frontend/src/App.css
@@ -1,5 +1,9 @@
 :root {
   --blue: #4c8bf5;
+  --green: #34a853;
+  --red: #ea4335;
+  --yellow: #fbbc05;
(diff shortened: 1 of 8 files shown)


frontend/src/App.css
```diff
@@ -1,5 +1,9 @@
 :root {
   --blue: #4c8bf5;
+  --green: #34a853;
+  --red: #ea4335;
+  --yellow: #fbbc05;
+  --gray: #6c757d;
```
genedwards, line 6: 👍

PR #157, opened 2020-07-16
show attraction name on hover/selected

Cindyzhang977:
Show attraction name when an attraction is hovered or selected

Left: hover
Right: selected
image


zoebarth:
Why did you decide to align text on the top right? Perhaps aligning left would look better?

PR #171, opened 2020-07-17
User & Trip CRU Utils

RCAVelez:
**What's new:**
Fetch request functions to create and read user

### DIFF
--- a/frontend/src/TripMock.js
+++ b/frontend/src/TripMock.js
@@ -1,0 +1,33 @@
+export const tripData = {
+  isOptimized: true,
+  searchText: 'Milano',
+  tripName: 'My Milan Trip',
+  attractions: [
+    {
+      attractionName: 'Milano Giuseppe',
+      photoReference: '2234f23f23r133fqfqef',
+      routeIndex: 0,
+      coordinates: {
+        lat: 1,
+        lng: 1,
+      },
+    },
+  ],
+};
+
+export const tripData2 = {
+  isOptimized: true,
+  searchText: 'Milano',
+  tripName: 'My Crazy Milan Trip',
+  attractions: [
+    {
+      attractionName: 'Milano Giuseppe',
+      photoReference: '2234f23f23r133fqfqef',
+      routeIndex: 0,
+      coordinates: {
+        lat: 1,
+        lng: 1,
+      },
+    },
+  ],
+};
--- a/frontend/src/TripMock.js
+++ b/frontend/src/TripMock.js
@@ -1,4 +1,4 @@
-export const tripData = {
+export const TRIP_DATA = {
   isOptimized: true,
   searchText: 'Milano',
   tripName: 'My Milan Trip',
@@ -15,7 +15,7 @@ export const tripData = {
   ],
 };
 
-export const tripData2 = {
+export const TRIP_DATA_2 = {
   isOptimized: true,
   searchText: 'Milano',
   tripName: 'My Crazy Milan Trip',
--- a/frontend/src/TripUtils.js
+++ b/frontend/src/TripUtils.js
@@ -2,7 +2,7 @@ import { fetchJson } from './fetchJson.js';
 
 export const createTrip = async (email, tripData) => {
   const url = `/api/v1/createTrip?email=${email}&tripData=${JSON.stringify(tripData)}`;
-  return await fetchJson(url);
+  await fetch(url, { method: 'POST' });
 };
 
 export const getTrip = async (tripId) => {
@@ -14,7 +14,7 @@ export const updateTrip = async (tripId, tripData) => {
   const url = `/api/v1/updateTrip?tripId=${tripId}&tripData=${JSON.stringify(
     tripData
   )}`;
-  return await fetchJson(url);
+  await fetch(url, { method: 'POST' });
 };
 
 export const getAllUserTrips = async (email) => {
--- a/frontend/src/TripUtils.js
+++ b/frontend/src/TripUtils.js
@@ -1,0 +1,27 @@
+import { fetchJson } from './fetchJson.js';
+
+export const createTrip = async (email, tripData) => {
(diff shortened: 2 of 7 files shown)


frontend/src/UserCRUD.js
```diff
@@ -0,0 +1,44 @@
+export const createUser = (email) => {
```
zoebarth, line 1: It looks like a lot of copy paste here, so rather than having three different methods, you should have one method that takes in a parameter and fetches the appropriate URL based on the parameter.

Cindyzhang977: +1

frontend/src/UserCRUD.js
```diff
@@ -0,0 +1,44 @@
...
+      if (response.status !== 200) {
+        return {};
+      }
+      response.json().then((response) => {
+        return response;
+      });
+    })
+    .catch((err) => {
+      return {};
+    });
```
Cindyzhang977, line 28: Can you write something like this, where you check that the response is 200 before asking for the json. Or I think you can also do `if (response.ok)`
```
  fetch(`/api/v1/readUser?email=${email}`)
    .then((response) => {
      if (response.status === 200) {
        response.json().then((response) => {
          return response;
        }
      })
    .catch((err) => {
      return {};
    });
```

frontend/src/UserCRUD.js
```diff
@@ -0,0 +1,44 @@
+export const createUser = (email) => {
+  fetch(`/api/v1/createUser?email=${email}`)
+    .then((response) => {
+      if (response.status !== 200) {
+        return {};
+      }
+      response.json().then((response) => {
```
genedwards, line 7: Are we planning on doing anything with this response when it succeeds?
I'm asking because calling createUser currently returns nothing.
If you plan on doing something when it succeeds, you need to do something like
```JS
export const createUser = (email) => {
  return fetch(`/api/v1/createUser?email=${email}`)
    .then((response) => {
      if (response.status === 200) {
        return response.json().then((response) => {
          return response;
        })
      }
    });
});

// or using async/await
export const createUser = async (email) => {
  const response = await fetch(`/api/v1/createUser?email=${email}`);
  const json = await response.json();
  return json;
});
```

Same for the other functions.

RCAVelez: Done

frontend/src/UserCRUD.js
RCAVelez: not using this anymore

frontend/src/UserCRUD.js
```diff
@@ -0,0 +1,44 @@
+export const createUser = (email) => {
+  fetch(`/api/v1/createUser?email=${email}`)
+    .then((response) => {
+      if (response.status !== 200) {
```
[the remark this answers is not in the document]
RCAVelez: 200 means response ok. This PR doesnt use this anymore

frontend/src/UserCRUD.js
```diff
@@ -0,0 +1,17 @@
+export const createUser = async (email) => {
```
Cindyzhang977, line 1: Could you make a generic function to call for each of these in order to avoid copy pasting code? You can pass in email and the url.

zoebarth: +1

RCAVelez: no need for email just pass the url then to a fetchRequest function

frontend/src/TripUtils.js
```diff
@@ -0,0 +1,27 @@
+import { fetchJson } from './fetchJson.js';
+
+export const createTrip = async (email, tripData) => {
```
zoebarth, line 3: I think these methods can still all be combined into one method

frontend/src/fetchJson.js
```diff
@@ -0,0 +1,5 @@
+export const fetchJson = async (url) => {
```
zoebarth, line 1: I don't see the point in this method. Why not combine the other methods in TripUtils and just include these lines?

frontend/src/UserUtils.js
```diff
@@ -0,0 +1,13 @@
+import { fetchJson } from './fetchJson.js';
+
+export const createUser = async (email) => {
```
zoebarth, line 3: These can also be combined.

frontend/src/TripMock.js
```diff
@@ -0,0 +1,33 @@
+export const tripData = {
```
mecolmg, line 1: nit: constant variables should be UPPER_SNAKE_CASE
https://google.github.io/styleguide/jsguide.html#file-goog-module-exports

frontend/src/TripUtils.js
mecolmg: I like it as is, fetchJson is encapsulating a the common logic. This could be written a little more concisely though:
```js
export const createTrip = async (email, tripData) => 
    await fetchJson(`/api/v1/createTrip?email=${email}&tripData=${JSON.stringify(tripData)}`);
```

or at the minimum, it's not worth storing `json` in a variable just to return it the next line when you could just do
```js
return await fetchJson(url);
```

frontend/src/fetchJson.js
RCAVelez: Two reasons to keep all the methods as is
1) The logic in fetchJson is shared between all the methods, no need to repeat the logic.
2) If the methods were combined, documentation would need to be provided for how to build each url for the Servlets.

frontend/src/TripMock.js
RCAVelez: done

frontend/src/TripUtils.js
RCAVelez: I pushed changes using `return await fetchJson(url);` 
I'm keeping the url const since it's more readable than making it all into one line.

PR #172, opened 2020-07-17
Create User Servlet

RCAVelez:
**What's new:**
Create User Servlet

### DIFF
--- a/backend/src/main/java/com/google/sps/servlets/createUserServlet.java
+++ b/backend/src/main/java/com/google/sps/servlets/createUserServlet.java
@@ -1,0 +1,51 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.servlets;
+
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.gson.*;
+import com.google.sps.UserCrud;
+import java.io.IOException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that creates a new user */
+@WebServlet("/api/v1/createUser")
+public class createUserServlet extends HttpServlet {
+  DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+  UserCrud userCrud = new UserCrud(datastore);
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String email = request.getParameter("email");
+    if (email == "" || email == null) {
+      return;
+    }
+    if (userCrud.readUser(email) != null) {
+      return;
+    }
+    try {
+      userCrud.createUser(email);
+    } catch (Exception e) {
+      return;
+    }
+    JsonObject jsonResults = new JsonObject();
+    response.setContentType("application/json;");
+    response.getWriter().println(jsonResults);
+  }
+}
--- a/backend/src/main/java/com/google/sps/servlets/createUserServlet.java
+++ b/backend/src/main/java/com/google/sps/servlets/createUserServlet.java
@@ -14,11 +14,12 @@
 
 package com.google.sps.servlets;
 
-import com.google.appengine.api.datastore.DatastoreService;
-import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.gson.Gson;
 import com.google.gson.JsonObject;
 import com.google.sps.UserCrud;
 import java.io.IOException;
+import java.util.ArrayList;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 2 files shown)


backend/src/main/java/com/google/sps/servlets/createUserServlet.java
```diff
@@ -0,0 +1,51 @@
...
+@WebServlet("/api/v1/createUser")
+public class createUserServlet extends HttpServlet {
+  DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+  UserCrud userCrud = new UserCrud(datastore);
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String email = request.getParameter("email");
+    if (email == "" || email == null) {
+      return;
```
zoebarth, line 37: Why are you returning in a void method?

Cindyzhang977: I think `return;` is the same as "return void"

backend/src/main/java/com/google/sps/servlets/createUserServlet.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+    if (userCrud.readUser(email) != null) {
```
Cindyzhang977, line 39:
```suggestion
    if (userCrud.readUser(email)) {
```

backend/src/main/java/com/google/sps/servlets/createUserServlet.java
```diff
@@ -0,0 +1,51 @@
...
+      return;
+    }
+    try {
+      userCrud.createUser(email);
+    } catch (Exception e) {
+      return;
+    }
```
Cindyzhang977, line 46: Instead of try/catch and then returning in the catch, you can make this an `if(condition) { create user }`

backend/src/main/java/com/google/sps/servlets/createUserServlet.java
Cindyzhang977, line 39: Since this if does the same as the first if statement, you can combine the conditions into one. 
```
if (email == "" || email == null || userCrud.readUser(email)) { return; }
```

backend/src/main/java/com/google/sps/servlets/createUserServlet.java
```diff
@@ -0,0 +1,51 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.servlets;
+
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.gson.*;
```
Cindyzhang977, line 19:
```suggestion
import com.google.gson.JsonObject;
```

backend/src/main/java/com/google/sps/servlets/createUserServlet.java
Cindyzhang977: Also, createUser wouldn't error if an email is provided right? And you checked the email at the beginning.

backend/src/main/java/com/google/sps/servlets/createUserServlet.java
```diff
@@ -0,0 +1,51 @@
...
+public class createUserServlet extends HttpServlet {
+  DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+  UserCrud userCrud = new UserCrud(datastore);
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String email = request.getParameter("email");
+    if (email == "" || email == null) {
+      return;
+    }
```
mecolmg, line 38: Just returning here will break stuff and make it more difficult to debug. Consider throwing an [`IllegalArgumentException`](https://docs.oracle.com/javase/7/docs/api/java/lang/IllegalArgumentException.html) with a helpful error message instead.

backend/src/main/java/com/google/sps/servlets/createUserServlet.java
mecolmg: Oh my gosh, I didn't know you could suggest a change...

backend/src/main/java/com/google/sps/servlets/createUserServlet.java
```diff
@@ -0,0 +1,51 @@
...
+    if (userCrud.readUser(email) != null) {
+      return;
```
mecolmg, line 40: Once again, consider throwing an error instead. Just returning is going to make this a pain to debug.

backend/src/main/java/com/google/sps/servlets/createUserServlet.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+    try {
+      userCrud.createUser(email);
+    } catch (Exception e) {
+      return;
```
mecolmg, line 45: Same here. Returning will just make it hard to debug.

backend/src/main/java/com/google/sps/servlets/createUserServlet.java
RCAVelez: We return so the rest of the function isn't used

backend/src/main/java/com/google/sps/servlets/createUserServlet.java
RCAVelez: Oh neat!

backend/src/main/java/com/google/sps/servlets/createUserServlet.java
RCAVelez: removed try catch since it's not needed

backend/src/main/java/com/google/sps/servlets/createUserServlet.java
RCAVelez: great! Updated

backend/src/main/java/com/google/sps/servlets/createUserServlet.java
RCAVelez: done

backend/src/main/java/com/google/sps/servlets/createUserServlet.java
RCAVelez: Now throwing IllegalArgumentException

backend/src/main/java/com/google/sps/servlets/createUserServlet.java
RCAVelez: I combined these with the top if statement which now throws an IllegalArgumentException error

backend/src/main/java/com/google/sps/servlets/createUserServlet.java
RCAVelez: removed try catch

backend/src/main/java/com/google/sps/servlets/createUserServlet.java
```diff
@@ -0,0 +1,46 @@
...
+/** Servlet that creates a new user */
+@WebServlet("/api/v1/createUser")
+public class createUserServlet extends HttpServlet {
+  DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+  UserCrud userCrud = new UserCrud(datastore);
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String email = request.getParameter("email");
+    if (email == "" || email == null || userCrud.readEntity("email", email, "User") != null) {
```
Cindyzhang977, line 36: nit
```suggestion
    if (email == "" || email == null || userCrud.readEntity("email", email, "User")) {
```

RCAVelez: ` bad operand type com.google.appengine.api.datastore.Entity for unary operator '!'`
Keeping this way for now

backend/src/main/java/com/google/sps/servlets/createUserServlet.java
```diff
@@ -0,0 +1,46 @@
...
+      throw new IllegalArgumentException("Email passed is not valid");
+    }
+
+    userCrud.createUser(email);
+
+    JsonObject jsonResults = new JsonObject();
+    response.setContentType("application/json;");
+    response.getWriter().println(jsonResults);
```
mecolmg, line 44: Do you even need to do this?

PR #173, opened 2020-07-17
Read User Servlet

RCAVelez:
**What's new:**
Read user servlet


backend/src/main/java/com/google/sps/servlets/readUserServlet.java
```diff
@@ -0,0 +1,50 @@
...
+@WebServlet("/api/v1/readUser")
+public class readUserServlet extends HttpServlet {
+  DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+  UserCrud userCrud = new UserCrud(datastore);
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String email = request.getParameter("email");
+    if (email == "" || email == null) {
+      return;
```
zoebarth, line 38: Why are you returning in a void method?

backend/src/main/java/com/google/sps/servlets/readUserServlet.java
```diff
@@ -0,0 +1,50 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.servlets;
+
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.gson.*;
```
Cindyzhang977, line 20:
```suggestion
import com.google.gson.JsonObject;
```

backend/src/main/java/com/google/sps/servlets/readUserServlet.java
```diff
@@ -0,0 +1,50 @@
...
+  UserCrud userCrud = new UserCrud(datastore);
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String email = request.getParameter("email");
+    if (email == "" || email == null) {
+      return;
+    }
+    Entity userEntity = userCrud.readUser(email);
+    if (userEntity == null) {
```
Cindyzhang977, line 41:
```suggestion
    if (userEntity) {
```

Cindyzhang977: You can add this to the first if statement since they both just return.
```
if (email == "" || email == null || !userCrud.readUser(email))
```

backend/src/main/java/com/google/sps/servlets/readUserServlet.java
RCAVelez: now throwing IllegalArgumentException

backend/src/main/java/com/google/sps/servlets/readUserServlet.java
RCAVelez: done

backend/src/main/java/com/google/sps/servlets/readUserServlet.java
RCAVelez: Check out how it looks now.

backend/src/main/java/com/google/sps/servlets/readUserServlet.java
```diff
@@ -0,0 +1,50 @@
...
+    if (email == "" || email == null) {
+      throw new IllegalArgumentException("Email passed is not valid");
+    }
+    Entity userEntity = userCrud.readEntity("email", email, "User");
+    if (userEntity == null) {
+      throw new IllegalArgumentException("Email passed is not linked to user");
+    }
+    JsonObject jsonResults = new JsonObject();
+    jsonResults.addProperty("email", userEntity.getProperty("email").toString());
+    jsonResults.addProperty("trips", userEntity.getProperty("trips").toString());
```
mecolmg, line 46: Right now I think you're storing tripIds, so this will need to change.

frontend/src/UserCRUD.js
```diff
@@ -1,17 +1,5 @@
...
-
-export const readUser = async (email) => {
-  const response = await fetch(`/api/v1/readUser?email=${email}`);
-  const json = await response.json();
-  return json;
-};
-
-export const updateUser = async (email, trips) => {
-  const response = await fetch(`/api/v1/updateUser?email=${email}&trips=${trips}`);
+export const fetchRequest = async (url) => {
```
mecolmg, line 1: I don't think this belongs in this PR.

backend/src/main/java/com/google/sps/servlets/readUserServlet.java
```diff
@@ -0,0 +1,51 @@
...
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that returns the user data */
+@WebServlet("/api/v1/readUser")
+public class readUserServlet extends HttpServlet {
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
zoebarth, line 32: This looks a little bit copy pasted from create user servlet. Consider combining those classes/methods

mecolmg: Looks like the only thing duplicated is the request checks. I think this is fine.

PR #174, opened 2020-07-17
Make route page responsive

zoebarth:
Add styling to route page to make it more responsive to both large and small screens. Remove description from html, but keep infrastructure in place to add it back later.

Screenshot 2020-07-17 at 4 12 26 PM - Display 1
image


frontend/src/RouteView.module.css
```diff
@@ -1,6 +1,9 @@
 .routeListContainer {
+  margin: 0 20px;
   max-height: 70vh;
   overflow: auto;
+  justify-content: center;
+  padding-left: 15px;
```
Cindyzhang977, line 6: alphabetize

frontend/src/RouteView.module.css
```diff
@@ -49,3 +52,15 @@
 .modalText {
   margin-bottom: 10px;
 }
+
+.container {
+  width: 100%;
+}
+
+.row {
+  margin: 0px;
```
Cindyzhang977, line 61: nit: remove `px`

PR #177, opened 2020-07-17
Search bar bugs

Cindyzhang977:
Address search bar bug.

- on enter key press, search based on the first suggested autocomplete 
- only set predictions if there are matches
  - prevents predictions from being set to `null` and causing null pointer exception
- remove background color of matched text and set text to bold

image


frontend/src/search/Search.module.css
```diff
@@ -60,3 +60,8 @@
   top: 12px;
   z-index: 10;
 }
+
+mark {
```
zoebarth, line 64: Who's Mark? :)

Cindyzhang977: It's the element tag of the matched text. I couldn't figure out another way to add a style to `mark` since I could only find this when I inspect element.

frontend/src/search/Search.js
```diff
@@ -23,10 +23,28 @@ function Search() {
...
+        if (status === 'OK') {
+          setPredictions(newPredictions);
+        }
+      }
     );
   };
 
+  const handleSearch = (text) => {
+    if (text) {
+      const searchText = text.constrctor === Array ? text[0] : text;
```
mecolmg, line 36: nit: constructor is spelled wrong, and perhaps you shouldn't have two possible input types. Instead, you could just pass options to handleSearch in the handleOnKeyDown method.

frontend/src/search/Search.module.css
mecolmg: You could also just pass the prop [`renderMenuItemChildren`](https://github.com/ericgio/react-bootstrap-typeahead/blob/master/docs/Props.md#typeahead) to not use a `mark`, which would avoid having to define a global style. 

Here's where they define the default implementation: https://github.com/ericgio/react-bootstrap-typeahead/blob/19c98c2d12a976312d546db146e8e59040690e28/src/components/TypeaheadMenu.js#L44-L52

PR #183, opened 2020-07-21
Route View CSS

Cindyzhang977:
Make the route view more responsive between chromebook and monitor screen sizes. Move `TripName` above the list of routes. 

image

### DIFF
--- a/frontend/src/route/LocationCard.module.css
+++ b/frontend/src/route/LocationCard.module.css
@@ -5,7 +5,7 @@
 
 .locationCard,
 .locationImg {
-  height: 22vh;
+  height: 230px;
   object-fit: cover;
 }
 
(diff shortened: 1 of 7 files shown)


frontend/src/route/LocationCard.module.css
```diff
@@ -29,3 +29,19 @@
 .locationCardContainer:hover .arrows {
   opacity: 0.6;
 }
+
+.locationName {
+  margin-top: 6%;
```
zoebarth, line 34: Why did you choose to use 6% as opposed to pixels?

zoebarth: And what is the percentage referring to?

Cindyzhang977: From what I understand it's like a percentage of the total space the div fits in. I think using percentage allows it to be more responsive since the card can have different pixel heights on different screens.

PR #184, opened 2020-07-21
make number of images over 8 unselectable

zoebarth:
image

After the user selects 8 images, make it impossible to select any more, and alert the user.

### DIFF
--- a/frontend/src/ExploreView.js
+++ b/frontend/src/ExploreView.js
@@ -96,7 +96,7 @@ function Explore() {
             ) : (
               initialAttractions.map((attraction, index) => (
                 <Card
-                  className={`${styles.attractionContainer} ${
+                  className={`${styles.attractionContainer}${
                     attraction.selected
                       ? styles.selectedImage
                       : selectedAttractions.length < 8
--- a/frontend/src/ExploreView.js
+++ b/frontend/src/ExploreView.js
@@ -85,6 +85,9 @@ function Explore() {
     <Container className={styles.exploreContainer}>
       <Row>
         <Col sm={6}>
+          {selectedAttractions.length < 8 || (
+            <p className={styles.p}>You may select up to 8 attractions.</p>
+          )}
           <div className={styles.attractionImagesContainer}>
             {initialAttractions.length === 0 ? (
               <div className={styles.fillerText}>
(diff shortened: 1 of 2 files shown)


frontend/src/ExploreView.module.css
```diff
@@ -77,6 +77,17 @@
...
+
+.unselectable {
+  opacity: 0.3;
+  pointer-events: none;
+}
+
+.p {
+  height: 25px;
+  margin: -25px 0 0;
+  text-align: center;
```
Cindyzhang977, line 92: nit: can you make the margin a little more between the text and the pictures? Also could you make the text color `var(--gray)` (the gray set in `App.css`) so the contrast with the white background isn't too much?

frontend/src/ExploreView.js
```diff
@@ -94,7 +97,11 @@ function Explore() {
               initialAttractions.map((attraction, index) => (
                 <Card
                   className={`${styles.attractionContainer} ${
-                    attraction.selected ? styles.selectedImage : ''
+                    attraction.selected
+                      ? styles.selectedImage
+                      : selectedAttractions.length < 8
+                      ? ''
+                      : styles.unselectable
```
mecolmg, line 104: nit: this is hard to read and understand. Also, you're possibly adding an extra space unnecessarily. Maybe something like the following, or consider using something like the [`classnames package`](https://www.npmjs.com/package/classnames#alternate-bind-version-for-css-modules-)

```js
className={`${styles.attractionContainer}${
    attraction.selected ? ` ${styles.selectedImage}` : ''
  }${
    selectedAttractions.length >= 8 ? ` ${styles.unselectable}` : ''
  }`}
```

frontend/src/ExploreView.module.css
zoebarth: Done

frontend/src/ExploreView.js
zoebarth: The problem with what you wrote is that if the attraction is selected, I don't want it to be unselectable

zoebarth: The structure i wrote should mimic:
if (selected) {
} else {
  if(length < 8) {
  } else {
  }
}

PR #185, opened 2020-07-21
Lazy load pagination

Cindyzhang977:
Load 20 more images each time user scrolls to the bottom of the attractions (loads up to the 60 provided by the API).

- use `pagination` object returned by `placesService.nearbySearch` 
  - ref: https://developers.google.com/maps/documentation/javascript/examples/place-search-pagination
  - while `hasNextPage`, `getNextPage` is a function that returns the next page of results, otherwise `getNextPage` is null
- rename `initialAttractions` to `attractions`
- use the `useReducer` hook to append new attractions to existing attractions as another page loads
- implement `handleScroll`, which detects when user scrolls to the bottom of the existing attractions and sets the loadMore boolean state accordingly 
  - ref: https://stackoverflow.com/questions/45585542/detecting-when-user-scrolls-to-bottom-of-div-with-react-js


frontend/src/ExploreView.js
```diff
@@ -63,13 +80,14 @@ function Explore() {
...
 
-    const handleNearbySearch = (results, status) => {
+    const handleNearbySearch = (results, status, pagination) => {
       if (status === google.maps.places.PlacesServiceStatus.OK) {
-        const newAllAttractions =
-          initialAttractions.length === 0
-            ? getAllAttractions(results)
-            : initialAttractions;
-        setInitialAttractions(newAllAttractions);
+        dispatch({ attractions: getAllAttractions(results) });
```
mecolmg, line 85: nit: Can you just do something like:

```js
// Above (near the top of the method).
const [attractions, setAttractions] = useState([]);
// Here (in handleNearbySearch).
setAttractions([...attractions, ...getAllAttractions(results)]);
```

I think the reducer is basically just doing the same thing, but it's a little harder to follow.

Cindyzhang977: I added more to the reducer when making the search radius slider. There's a lot more logic in the reducer now.

PR #187, opened 2020-07-22
Trip CRU class

RCAVelez:
**What's new :**
Trip Crud Class


backend/src/main/java/com/google/sps/TripCRUD.java
```diff
@@ -0,0 +1,138 @@
...
+import com.google.gson.JsonParser;
+import java.util.ArrayList;
+
+/** Class to handles CRU related to the Trip */
+public class TripCRUD {
+  private static final DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+  /** private constructor */
+  private TripCRUD(){};
+
```
Cindyzhang977, line 21: What is this used for?

backend/src/main/java/com/google/sps/TripCRUD.java
```diff
@@ -0,0 +1,138 @@
...
+   *
+   * @param tripId id of the trip to find
+   * @param tripData string representation of tripData json
+   */
+  public static void updateTrip(String tripId, String tripData) {
+    Entity tripEntity;
+    try {
+      tripEntity = TripCRUD.readTrip(tripId);
+    } catch (EntityNotFoundException e) {
+      return;
```
Cindyzhang977, line 133: should this return or throw an exception?

backend/src/main/java/com/google/sps/TripCRUD.java
RCAVelez: This is a private constructor will prevent the instantiation of this class directly. That's so the CRU logic is completely abstracted.

backend/src/main/java/com/google/sps/TripCRUD.java
RCAVelez: return because the method is not meant to throw an error.

backend/src/main/java/com/google/sps/TripCRUD.java
Cindyzhang977: But it's currently empty. I think we can just remove it.

RCAVelez: I think so. Let me see

backend/src/main/java/com/google/sps/TripCRUD.java
```diff
@@ -0,0 +1,138 @@
...
+
+  /**
+   * Converts Trip Entity to Json
+   *
+   * @param tripEntity Trip Entity
+   * @return JsonObject of trip
+   */
+  public static JsonObject toJson(Entity tripEntity) {
+    JsonObject jsonTrip = new JsonObject();
+    jsonTrip.addProperty("tripId", Long.toString(tripEntity.getKey().getId()));
```
mecolmg, line 100: In this specific case it makes sense to use a string because you're going to be reading this value in JavaScript which doesn't support Longs (it only supports up to 53-bit numbers due to floating point precision). 

So it's good as is, just wanted to point out why 👍

backend/src/main/java/com/google/sps/TripCRUD.java
```diff
@@ -0,0 +1,138 @@
...
+    jsonTrip.addProperty(
+        "isOptimized", Boolean.parseBoolean(tripEntity.getProperty("isOptimized").toString()));
```
mecolmg, line 102: optional: for some added safety, each of these could be checked for `tripEntity.hasProperty(property)` before accessing them.

mecolmg, line 102: Entity.getProperty returns an object which you can cast to the correct type, so here (and below) can just be
```suggestion
    jsonTrip.addProperty("isOptimized", (Boolean) tripEntity.getProperty("isOptimized"));
```

backend/src/main/java/com/google/sps/TripCRUD.java
```diff
@@ -0,0 +1,138 @@
...
+    jsonTrip.addProperty("searchText", tripEntity.getProperty("searchText").toString());
+    jsonTrip.addProperty("tripName", tripEntity.getProperty("tripName").toString());
+    ArrayList<JsonObject> attractions = new ArrayList<JsonObject>();
```
mecolmg, line 105: Instead of building an ArrayList, you should be building an [`JsonArray`](https://www.javadoc.io/static/com.google.code.gson/gson/2.8.5/com/google/gson/JsonArray.html) which you can add to `jsonTrip` with like:
```java
JsonArray attractions = new JsonArray();
for (EmbeddedEntity attraction :
    (ArrayList<EmbeddedEntity>) tripEntity.getProperty("attractions")) {
  JsonObject attractionJson = new JsonObject();
  // populate attractionJson.
  attractions.add(attractionJson);
}
jsonTrip.add("attractions", attractions);
```

backend/src/main/java/com/google/sps/TripCRUD.java
```diff
@@ -0,0 +1,138 @@
...
+    for (EmbeddedEntity attraction :
+        (ArrayList<EmbeddedEntity>) tripEntity.getProperty("attractions")) {
+      JsonObject attractionJson = new JsonObject();
+      attractionJson.addProperty(
+          "attractionName", attraction.getProperty("attractionName").toString());
```
mecolmg, line 110: Instead of `toString()` do `(String) attraction.getProperty(property)`. The object itself is already a string, you just need to tell it that it is.

backend/src/main/java/com/google/sps/TripCRUD.java
mecolmg: I think @Cindyzhang977 makes a good point. Throwing an exception here would be preferable, because otherwise the call to updateTrip can silently fail.

backend/src/main/java/com/google/sps/TripCRUD.java
```diff
@@ -0,0 +1,138 @@
...
+    return jsonTrip;
+  }
+
+  /**
+   * Updates a trip's properties
+   *
+   * @param tripId id of the trip to find
+   * @param tripData string representation of tripData json
+   */
+  public static void updateTrip(String tripId, String tripData) {
```
mecolmg, line 128: Possibly something that should be put in a future PR, but consider how the optimize servlet will update the routeIndex. With this implementation, it'd have to build manipulate a json object, serialize it, then call this method. Could that be improved as an API?

backend/src/main/java/com/google/sps/TripCRUD.java
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  /**
+   * Mutates trip entity properties given string representation of tripData json
+   *
+   * @param tripEntity Trip Entity
+   * @param tripData tripData string representation of a trip json
+   */
+  private static void setProperties(Entity tripEntity, String tripData) {
+    JsonParser parser = new JsonParser();
```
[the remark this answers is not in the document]
RCAVelez: it is private static

backend/src/main/java/com/google/sps/TripCRUD.java
RCAVelez: yay!

backend/src/main/java/com/google/sps/TripCRUD.java
```diff
@@ -0,0 +1,138 @@
...
+    JsonElement jsonElement = parser.parse(tripData);
+    JsonObject jsonObject = jsonElement.getAsJsonObject();
+
+    tripEntity.setProperty(
+        "isOptimized", Boolean.parseBoolean(jsonObject.get("isOptimized").toString()));
```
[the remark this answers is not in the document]
RCAVelez: done

backend/src/main/java/com/google/sps/TripCRUD.java
RCAVelez: okidoki, method now throws EntityNotFoundException

backend/src/main/java/com/google/sps/TripCRUD.java
RCAVelez: okidoki done

backend/src/main/java/com/google/sps/TripCRUD.java
RCAVelez: done

backend/src/main/java/com/google/sps/TripCRUD.java
RCAVelez: I think that's good. We can make an issue out of it while the Trip Class is functional and the tests are made

backend/src/main/java/com/google/sps/TripCRUD.java
RCAVelez: done

backend/src/main/java/com/google/sps/TripCRUD.java
RCAVelez: done

backend/src/main/java/com/google/sps/servlets/createUserServlet.java
```diff
@@ -0,0 +1,36 @@
...
+import com.google.sps.UserCrud;
+import java.io.IOException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that creates a new user */
+@WebServlet("/api/v1/createUser")
+public class createUserServlet extends HttpServlet {
```
mecolmg, line 26: Oh woops, please fix! All classes should be `UpperCamelCase`!

RCAVelez: Done

backend/src/main/java/com/google/sps/servlets/createUserServlet.java
```diff
@@ -0,0 +1,36 @@
+// Copyright 2019 Google LLC
```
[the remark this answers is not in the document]
RCAVelez: yes, the base was set to this branch instead of master and it was merged accidentally.

frontend/src/TripUtils.js
```diff
@@ -1,4 +1,4 @@
-import { fetchJson } from './fetchJson.js';
+import { fetchJson } from './FetchJson.js';
```
mecolmg, line 1: super minor nit: Generally in react you only UpperCamelCase component names. Utility files like `fetchJson` are either lowerCamelCase or snake_case depending on your codebase (I'd just keep it as it was as lowerCamelCase).

RCAVelez: done

PR #189, opened 2020-07-22
Create Trip Servlet

RCAVelez:


backend/src/main/java/com/google/sps/servlets/createTripServlet.java
```diff
@@ -0,0 +1,56 @@
...
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet that creates a new trip */
+@WebServlet("/api/v1/createTrip")
+public class createTripServlet extends HttpServlet {
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String email = request.getParameter("email");
+
+    if (email == "" || email == null) {
```
zoebarth, line 34: This also looks copy pasted

Cindyzhang977: I feel like this is ok since different errors are thrown

backend/src/main/java/com/google/sps/servlets/createTripServlet.java
```diff
@@ -0,0 +1,56 @@
...
+      throw new IllegalArgumentException("Email passed is not linked to user");
+    }
+
+    Entity tripEntity = TripCRUD.createTrip(email, tripData);
+
+    JsonObject jsonResults = new JsonObject();
+    jsonResults.addProperty("tripId", tripEntity.getKey().getId());
+    jsonResults.addProperty("tripData", tripData);
+    response.setContentType("application/json;");
+    response.getWriter().println(jsonResults);
```
Cindyzhang977, line 54: Since this is creating the trip and putting it in datastore, does it still need to return a json object? This method should also probably be a `goPost` and not return anything.

backend/src/main/java/com/google/sps/servlets/createTripServlet.java
mecolmg: +1 to it's okay

PR #190, opened 2020-07-22
Read Trip Servlet

RCAVelez:


backend/src/main/java/com/google/sps/servlets/readTripServlet.java
```diff
@@ -0,0 +1,51 @@
...
+
+    if (tripId == "" || tripId == null) {
+      throw new IllegalArgumentException("tripId passed is not valid");
+    }
+
+    Entity tripEntity;
+    try {
+      tripEntity = TripCRUD.readTrip(tripId);
+    } catch (EntityNotFoundException e) {
+      return;
```
Cindyzhang977, line 42: Should this return or throw an error?

backend/src/main/java/com/google/sps/servlets/readTripServlet.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+    JsonObject tripJson = TripCRUD.toJson(tripEntity);
+    JsonObject jsonResults = new JsonObject();
+    jsonResults.addProperty("tripId", tripId);
+    jsonResults.addProperty("tripData", tripJson.toString());
+    response.setContentType("application/json;");
+    response.getWriter().println(jsonResults);
```
Cindyzhang977, line 49: What's the purpose of creating a new `JsonObject` `jsonResults`? Doesn't the `toJson` function already return a `JsonObject` that has the `tripId` property and everything that would go inside `tripData` (`tripName`, `searchText`, etc.)? This servlet could directly return `tripJson`.

Cindyzhang977: I noticed that in `TripCRUD.createTrip` you call `toEntity` which calls `setProperties`. Since each property in `tripData` is being added to the entity individaully, there's no need to maintain the `tripData` property in a json object. If I understand correctly, your code currently return a json object like:
```
{
  isOptimized: false,
  tripName: "",
  searchText: "",
  attractions: [],
  tripId: "",
  tripData: {
    isOptimized: false,
    tripName: "",
    searchText: "",
    attractions: [],
    tripId: "",
  }
}
```

RCAVelez: I see how it's redundant to recreate a jsonResults. I can directly return from the .toJson method

This is the current response.
```
{
  tripId: "",
  tripData: {
    isOptimized: false,
    tripName: "",
    searchText: "",
    attractions: [],
    tripId: "",
  }
}
```

removing the id the response is then
`{
    isOptimized: false,
    tripName: "",
    searchText: "",
    attractions: [],
    tripId: "",
  }`

backend/src/main/java/com/google/sps/servlets/readTripServlet.java
RCAVelez: now throwing illegalArumentException

backend/src/main/java/com/google/sps/servlets/readTripServlet.java
```diff
@@ -0,0 +1,47 @@
...
+
+    Entity tripEntity;
+    try {
+      tripEntity = TripCRUD.readTrip(tripId);
+    } catch (EntityNotFoundException e) {
+      throw new IllegalArgumentException("Trip not found");
+    }
+
+    response.setContentType("application/json;");
+    response.getWriter().println(TripCRUD.toJson(tripEntity));
```
mecolmg, line 45: Oh wait, TripCRUD.toJson returns a JsonObject, do you need to call .toString()?

RCAVelez: yup, done

backend/src/main/java/com/google/sps/servlets/readTripServlet.java
```diff
@@ -0,0 +1,47 @@
...
+
+    if (tripId == "" || tripId == null) {
+      throw new IllegalArgumentException("tripId passed is not valid");
+    }
+
+    Entity tripEntity;
+    try {
+      tripEntity = TripCRUD.readTrip(tripId);
+    } catch (EntityNotFoundException e) {
+      throw new IllegalArgumentException("Trip not found");
```
mecolmg, line 41: This isn't really an `IllegalArgumentException`, can you just let the `EntityNotFoundException` throw?

PR #191, opened 2020-07-22
Update Trip Servlet

RCAVelez:


backend/src/main/java/com/google/sps/servlets/updateTripServlet.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    TripCRUD.updateTrip(tripId, tripData);
+
+    JsonObject jsonResults = new JsonObject();
+    jsonResults.addProperty("tripId", tripId);
+    jsonResults.addProperty("tripData", tripData);
+
+    response.setContentType("application/json;");
+    response.getWriter().println(jsonResults);
```
Cindyzhang977, line 44: I think this servlet should also be a `doPost` like `\createTrip`, since you're not retrieving a trip from the database to return to the webpage. Instead, this servlet is used when the user clicks save after editing an exiting page, meaning that the frontend has the up-to-date information already and it's only the backend that needs to be updated.

RCAVelez:
Attraction and Attraction Panel delete was in a different PR but they were merged in by one of us...

backend/src/main/java/com/google/sps/servlets/updateTripServlet.java
RCAVelez: changed to doPost,done

PR #196, opened 2020-07-22
Search radius slider

Cindyzhang977:
Adjust search radius for nearby places using a slider.

- display name of location currently searching near
- include slider bar for user to adjust the radius to search nearby places
- fix lazy load bug where pagetoken resets from beginning once any image is clicked

image

KNOWN BUG: pagetoken resets when an image that is in the second lazily loaded batch is clicked. This causes the second batch of images to be loaded again.


frontend/src/ExploreView.js
```diff
@@ -36,82 +37,166 @@ function Explore() {
...
+  );
+
+  const handleNearbySearch = useCallback(
+    (status, pagination) => {
+      if (status === 'OK' && getNextPage.current !== 'end') {
+        dispatch({
+          attractions: getAllAttractions(nearbySearchResults.current),
+          radius: radiusState,
+        });
+        getNextPage.current = pagination.hasNextPage
```
genedwards, line 109: This seems quite hacky. getNextPage representing both a function and a string doesn't seem right to me. What do you think about a `hasNextPage` boolean instead?

frontend/src/ExploreView.js
```diff
@@ -36,82 +37,166 @@ function Explore() {
...
+  const [state, dispatch] = useReducer(reducer, {
+    attractions: [],
+    radius: radiusState,
+  });
 
-  const [loadMore, setLoadMore] = useState(false);
-  const getNextPage = useRef(null);
-  const history = useHistory();
+  useEffect(() => {
+    tripObject.selectedAttractions = selectedAttractions;
```
genedwards, line 68: Shouldn't we update it as if it's immutable?
```JS
useEffect(() => {
  setTripObject({...tripObject, selectedAttractions});
```

frontend/src/ExploreView.js
```diff
@@ -36,82 +37,166 @@ function Explore() {
...
+          const photoUrl = attraction.photos[0].getUrl();
+          const latLng = attraction.geometry.location;
+          const isSelected = selectedAttractions.some(
+            (newAttraction) => newAttraction.photoUrl === attraction.photos[0].getUrl()
+          );
+          const newAttraction = createAttraction(name, latLng, photoUrl, isSelected);
+          newAllAttractions.push(newAttraction);
+        }
+      }
+      return newAllAttractions;
```
genedwards, line 97: this seems unnecessarily expensive, I think we can use attractions.map() and reduce it to O(n) time

frontend/src/ExploreView.js
Cindyzhang977: I initially set `hasNextPage` to false when it reached the last page, but I wanted something that's more distinguishable from `null` so I set it to `"end"`.

frontend/src/ExploreView.js
Cindyzhang977: oof this causes an infinite loop and somehow also breaks cloud shell editor. It's pretty hacky, but `tripObject` is set once from `null` to a json object and then mutated later on to prevent rerendering.

frontend/src/ExploreView.js
Cindyzhang977: Is this for loop O(n) either way, since we have to loop through everything? With the `if` condition in the for loop, if we use `map()` would we also have to use `filter()` before, so we end up looping through everything twice?

frontend/src/ExploreView.js
mecolmg: Could you just make it a no-op function (i.e. `() => {}`), in which case you don't even need to check anything

frontend/src/ExploreView.js
genedwards: ok, I see the infinite loop

frontend/src/ExploreView.js
```diff
@@ -36,82 +37,166 @@ function Explore() {
   const [selectedAttractions, setSelectedAttractions] = useState(
     tripObject.selectedAttractions
   );
+  const [radiusState, setRadiusState] = useState(8);
```
mecolmg, line 40: nit: can we just call it radius to be consistent, since none of the other state variables end in `State`

frontend/src/ExploreView.js
```diff
@@ -36,82 +37,166 @@ function Explore() {
...
-        getNextPage.current = pagination.hasNextPage
-          ? () => {
-              pagination.nextPage();
-            }
-          : null;
-      } else {
-        setLoading(false);
-      }
-    };
+  useEffect(() => {
```
mecolmg, line 150: Can you add a comment explaining what this does?

frontend/src/ExploreView.js
genedwards: oh, right, hmm
this is why I asked for selectedAttractions to be filtered from all attractions https://github.com/googleinterns/step88-2020/pull/102#discussion_r452892053

I don't know why we moved away from that, but I won't block you

frontend/src/index.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import 'bootstrap/dist/css/bootstrap.min.css';
+import 'react-bootstrap-range-slider/dist/react-bootstrap-range-slider.css';
```
mecolmg, line 4: Hmm, since the slider is only used in Explore, maybe import it there instead so that we only download this if necessary.

frontend/src/ExploreView.js
genedwards: I would rather have `getNextPage` be null when we reached the last page. I really dislike the thought of something being both a function and a string.

frontend/src/ExploreView.js
```diff
@@ -36,82 +37,166 @@ function Explore() {
...
+    if (textSearchResults.current) {
+      handleTextSearch('OK');
+    }
+  }, [radiusState]);
 
-    const placesService = new google.maps.places.PlacesService(map);
-    placesService.textSearch(
+  function handleScroll(e) {
+    const loadMoreFlag =
+      e.target.scrollHeight - e.target.scrollTop === e.target.clientHeight;
```
genedwards, line 158: Colm asked this to be >= before for safety on some odd browsers, let's keep it that way

frontend/src/ExploreView.js
```diff
@@ -36,82 +37,166 @@ function Explore() {
...
+  /**
+   * Get the photo url of each attraction object
+   * @param {object[]} attractions array of objects from Places Request
+   * @return {object[]} array of all attractions
+   */
+  const getAllAttractions = useCallback(
+    (attractions) => {
+      const newAllAttractions = [];
+      for (const attraction of attractions) {
+        if ('photos' in attraction) {
```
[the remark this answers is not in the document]
genedwards: maybe even `if (!!attraction.photos)`

frontend/src/ExploreView.js
Cindyzhang977: oops that PR didn't get merged in

PR #204, opened 2020-07-27
Trip tests

RCAVelez:


backend/src/test/java/com/google/sps/TripCRUDTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+  @Test
+  public void readTrip_returnsEntityForTrip() {
+    Entity userEntity = UserCrud.createUser(EMAIL);
+    Entity tripEntity = TripCRUD.createTrip(EMAIL, TRIP_DATA);
+    Entity readTripEntity = new Entity("Trip");
+    try {
+      readTripEntity = TripCRUD.readTrip(Long.toString(tripEntity.getKey().getId()));
+    } catch (EntityNotFoundException e) {
+    }
```
Cindyzhang977, line 73: Instead of try/catch in a junit test, you can explicitly state that an error is expected. I think something like `@Test(expected = EntityNotFoundException.class)` should work (might need some imports?). This would require you to split this test in two: one where an error is expected to be thrown and one where you `asserEquals`. [(Ref)](https://www.guru99.com/junit-exception-test.html#:~:text=JUnit%20provides%20the%20facility%20to,throwing%20expected%20exception%20or%20not.&text=While%20Testing%20exception%2C%20you%20need,test%20annotation%20is%20the%20same.)

backend/src/test/java/com/google/sps/TripCRUDTest.java
```diff
@@ -0,0 +1,103 @@
...
+  @Test
+  public void updateTrip_returnsUpdatedTripNameForUpdatedEntity() {
+    Entity userEntity = UserCrud.createUser(EMAIL);
+    Entity tripEntity = TripCRUD.createTrip(EMAIL, TRIP_DATA);
+    TripCRUD.updateTrip(Long.toString(tripEntity.getKey().getId()), TRIP_DATA_2);
+    Entity tripFound = new Entity("Trip");
+    try {
+      tripFound = TripCRUD.readTrip(Long.toString(tripEntity.getKey().getId()));
+    } catch (EntityNotFoundException e) {
+    }
```
Cindyzhang977, line 100: same thing for the try/catch here

.gitignore
```diff
@@ -191,6 +191,9 @@ dist
 **/*.backup.*
```
mecolmg, line 191: Can you fix your git branch to not show these and other unrelated changes?

backend/src/test/java/com/google/sps/TripCRUDTest.java
```diff
@@ -0,0 +1,106 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps;
+
+import static org.junit.Assert.assertEquals;
+
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.EntityNotFoundException;
```
mecolmg, line 20: nit: unused

backend/src/main/java/com/google/sps/TripCRUD.java
```diff
@@ -84,9 +84,15 @@ private static void setProperties(Entity tripEntity, String tripData) {
...
-  public static Entity readTrip(String tripId) throws EntityNotFoundException {
+  public static Entity readTrip(String tripId) {
     Key entityKey = KeyFactory.createKey("Trip", Long.parseLong(tripId));
-    return datastore.get(entityKey);
+    Entity tripEntity;
+    try {
+      tripEntity = datastore.get(entityKey);
+    } catch (EntityNotFoundException e) {
+      return null;
+    }
```
mecolmg, line 94: Did you change this for testing? In general, you shouldn't change functionality of what you're testing to make it easier to test. That's a codesmell.

To test exceptions in JUnit4, see: https://www.baeldung.com/junit-assert-exception#junit-4

backend/src/main/java/com/google/sps/servlets/readTripServlet.java
```diff
@@ -35,12 +35,11 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
     }
 
-    Entity tripEntity;
-    try {
-      tripEntity = TripCRUD.readTrip(tripId);
-    } catch (EntityNotFoundException e) {
-      return;
+    Entity tripEntity = TripCRUD.readTrip(tripId);
+    if (tripEntity == null) {
+      throw new IllegalArgumentException("trip not found");
```
mecolmg, line 40: nit:
```suggestion
      throw new IllegalArgumentException("Trip not found");
```

backend/src/test/java/com/google/sps/TripCRUDTest.java
```diff
@@ -0,0 +1,106 @@
...
+    Entity userEntity = UserCrud.createUser(EMAIL);
+    Entity tripEntity = TripCRUD.createTrip(EMAIL, TRIP_DATA);
+    Entity actual = TripCRUD.readTrip(INVALID_TRIP_ID);
+    assertEquals(null, actual);
+  }
+
+  @Test
+  public void toEntity_returnsTripEntityFromJsonMatchingTripName() {
+    Entity tripEntityConverted = TripCRUD.toEntity(TRIP_DATA, "", null);
+    assertEquals("\"My Milan Trip\"", (String) tripEntityConverted.getProperty("tripName"));
```
mecolmg, line 79: Hmm, should it have the extra quotes stored? Is this a bug?

backend/src/test/java/com/google/sps/TripCRUDTest.java
```diff
@@ -0,0 +1,106 @@
...
+    Entity tripEntity = TripCRUD.createTrip(EMAIL, TRIP_DATA);
+    TripCRUD.updateTrip(Long.toString(tripEntity.getKey().getId()), TRIP_DATA_2);
+    Entity tripFound = TripCRUD.readTrip(Long.toString(tripEntity.getKey().getId()));
+    assertEquals("\"My Awesome Milan Trip\"", (String) tripFound.getProperty("tripName"));
+  }
+
+  @Test
+  public void updateTrip_noTripFound() {
+    Entity userEntity = UserCrud.createUser(EMAIL);
+    Entity tripEntity = TripCRUD.createTrip(EMAIL, TRIP_DATA);
```
mecolmg, line 102: nit: not really necessary

backend/src/test/java/com/google/sps/TripCRUDTest.java
```diff
@@ -0,0 +1,106 @@
...
+    Entity actual = TripCRUD.readTrip(INVALID_TRIP_ID);
+    assertEquals(null, actual);
+  }
+}
```
mecolmg, line 106: nit: add edge case tests for:
- Creating a trip without creating a user first
- toJson with missing entity info
- toEntity with missing or malformed json
- Any other edge cases you can think of.

frontend/src/RouteView.js
```diff
@@ -60,6 +61,13 @@ function RouteView({ loggedIn }) {
   function save() {
     setIsSaved(true);
     // save to back end database
+    tripObject.isOptimized = isOptimized;
+    console.log(tripObject);
+    if (!tripObject.id) {
+      createTrip(userEmail, tripObject).then((tripId) => console.log(tripId));
+    } else {
+      updateTrip(tripObject.id, tripObject);
+    }
```
Cindyzhang977, line 70: please ignore this. it was accidentally pushed from another branch when trying to push java updates.

backend/src/test/java/com/google/sps/TripCRUDTest.java
zoebarth: Done

backend/src/main/java/com/google/sps/TripCrud.java
```diff
@@ -27,8 +26,21 @@
...
+  /**
+   * Add tripId property to Trip Entity
+   *
+   * @param id the trip id
+   */
+  public static void addTripId(String id) {
+    Entity tripEntity = TripCrud.readTrip(id);
+    tripEntity.setProperty("tripId", id);
+    datastore.put(tripEntity);
   }
```
mecolmg, line 44: You shouldn't need to store tripId as a property of the entity, since you can get it as entity.getKey().getId()

I think this method can be removed, and if for some reason it can't, it should be made private.

backend/src/main/java/com/google/sps/TripCrud.java
```diff
@@ -55,22 +67,21 @@ private static void setProperties(Entity tripEntity, String tripData) {
...
+      embeddedAttraction.setProperty("photoUrl", attraction.get("photoUrl").toString());
       embeddedAttraction.setProperty(
-          "attractionName", attraction.get("attractionName").getAsString());
-      embeddedAttraction.setProperty(
-          "photoReference", attraction.get("photoReference").getAsString());
-      embeddedAttraction.setProperty("routeIndex", attraction.get("routeIndex").getAsInt());
-      embeddedAttraction.setProperty("coordinates", attraction.get("coordinates").getAsString());
+          "routeIndex", Integer.parseInt(attraction.get("routeIndex").toString()));
+      embeddedAttraction.setProperty("lat", attraction.get("lat").toString());
+      embeddedAttraction.setProperty("lng", attraction.get("lng").toString());
```
mecolmg, line 84: nit: This seems like a regression, what was wrong with the existing code?

backend/src/main/java/com/google/sps/TripCrud.java
```diff
@@ -27,8 +26,21 @@
   public static Entity createTrip(String email, String tripData) {
     Entity tripEntity = toEntity(tripData, "", null);
     datastore.put(tripEntity);
-    UserCrud.addTripId(email, tripEntity.getKey().getId());
-    return tripEntity;
+    long id = tripEntity.getKey().getId();
+    UserCrud.addTripId(email, id);
+    TripCrud.addTripId(Long.toString(id));
+    return TripCrud.readTrip(Long.toString(id));
```
mecolmg, line 32: Why can't you just return tripEntity?

backend/src/main/java/com/google/sps/TripCrud.java
```diff
@@ -80,12 +91,17 @@ private static void setProperties(Entity tripEntity, String tripData) {
    * Finds a trip entity by id
    *
    * @param tripId id of the trip to find
-   * @throws EntityNotFoundException if trip entity is not found
-   * @return Trip entity
+   * @return Trip entity if found, null if EntityNotFoundException is caught
    */
-  public static Entity readTrip(String tripId) throws EntityNotFoundException {
+  public static Entity readTrip(String tripId) {
```
mecolmg, line 96: `entity.getKey().getId()` returns a long, so it would make more sense for this to take in a long.

backend/src/main/java/com/google/sps/TripCrud.java
```diff
@@ -97,23 +113,24 @@ public static Entity readTrip(String tripId) throws EntityNotFoundException {
...
-          "photoReference", (String) attraction.getProperty("photoReference"));
-      attractionJson.addProperty("routeIndex", (Integer) attraction.getProperty("routeIndex"));
-      attractionJson.addProperty("coordinates", (String) attraction.getProperty("coordinates"));
+          "routeIndex", Integer.parseInt(attraction.getProperty("routeIndex").toString()));
+      attractionJson.addProperty("lat", attraction.getProperty("lat").toString());
+      attractionJson.addProperty("lng", attraction.getProperty("lng").toString());
       attractions.add(attractionJson);
     }
-    jsonTrip.add("attractions", attractions);
+    jsonTrip.addProperty("attractions", attractions.toString());
```
mecolmg, line 133: Likewise, this seems like a regression. Why are you storing this as a string? This will mess with deserializing the object from the client.

The way it was implemented previously by storing a JsonArray using the `add` method seems like the correct way to do this: https://www.javadoc.io/static/com.google.code.gson/gson/2.6.2/com/google/gson/JsonObject.html#add-java.lang.String-com.google.gson.JsonElement-

backend/src/main/java/com/google/sps/TripCrud.java
```diff
@@ -124,10 +141,8 @@ public static JsonObject toJson(Entity tripEntity) {
    * @param tripData string representation of tripData json
    */
   public static void updateTrip(String tripId, String tripData) {
```
mecolmg, line 143: Likewise, tripId should probably be a long

backend/src/test/java/com/google/sps/TripCrudTest.java
```diff
@@ -0,0 +1,132 @@
...
+    assertEquals(tripEntity, readTripEntity);
+  }
+
+  @Test
+  public void readTrip_returnsEntityForTrip() {
+    Entity userEntity = UserCrud.createUser(EMAIL);
+    Entity tripEntity = TripCrud.createTrip(EMAIL, TRIP_DATA);
+    Entity readTripEntity = TripCrud.readTrip(Long.toString(tripEntity.getKey().getId()));
+    assertEquals(tripEntity, readTripEntity);
+  }
```
mecolmg, line 66: These two tests are the same

backend/src/test/java/com/google/sps/TripCrudTest.java
```diff
@@ -0,0 +1,132 @@
...
+  public void tearDown() {
+    DATASTORE_SERVICE_HELPER.tearDown();
+  }
+
+  @Test
+  public void createTrip_returnsEntityForCreatedTrip() {
+    Entity userEntity = UserCrud.createUser(EMAIL);
+    Entity tripEntity = TripCrud.createTrip(EMAIL, TRIP_DATA);
+    Entity readTripEntity = TripCrud.readTrip(Long.toString(tripEntity.getKey().getId()));
+    assertEquals(tripEntity, readTripEntity);
```
mecolmg, line 57: Instead of this, you should create an expected Entity and verify that it equals the actual entity, or you could verify that each of the properties is instantiated as expected.

backend/src/test/java/com/google/sps/TripCrudTest.java
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  @Test
+  public void readTrip_noTripFound() {
+    Entity userEntity = UserCrud.createUser(EMAIL);
+    Entity tripEntity = TripCrud.createTrip(EMAIL, TRIP_DATA);
+    Entity actual = TripCrud.readTrip(INVALID_TRIP_ID);
+    assertEquals(null, actual);
```
mecolmg, line 73: nit: use assertNull

backend/src/main/java/com/google/sps/TripCrud.java
```diff
@@ -27,8 +26,21 @@
...
+   *
+   * @param id the trip id
+   */
+  public static void addTripId(String id) {
+    Entity tripEntity = TripCrud.readTrip(id);
+    tripEntity.setProperty("tripId", id);
+    datastore.put(tripEntity);
   }
 
   /**
```
mecolmg, line 46: Doesn't need to be in this PR, but there should really be two toEntity methods. One that takes in tripData, and one that takes in the tripData, tripId, and tripKey. Having to pass in null or empty strings to a method is a **big** codesmell in Java.

backend/src/test/java/com/google/sps/TripCrudTest.java
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  @Test
+  public void toEntity_returnsTripEntityFromJsonMatchingTripName() {
+    Entity tripEntityConverted = TripCrud.toEntity(TRIP_DATA, "", null);
+    assertEquals("\"My Milan Trip\"", (String) tripEntityConverted.getProperty("tripName"));
```
mecolmg, line 79: What's up with the extra quotes? I wouldn't expect those to be stored.

backend/src/test/java/com/google/sps/TripCrudTest.java
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  @Test
+  public void toJson_returnsTripJsonFromEntityMatchingSearchText() {
+    Entity userEntity = UserCrud.createUser(EMAIL);
+    Entity tripEntity = TripCrud.createTrip(EMAIL, TRIP_DATA);
+    JsonObject tripDataJson = TripCrud.toJson(tripEntity);
+    assertEquals("\"Milano\"", tripDataJson.get("searchText").getAsString());
```
mecolmg, line 87: Same with extra quotes

backend/src/test/java/com/google/sps/TripCrudTest.java
mecolmg, line 79: Can we check more than just the trip name? Testing the entire entity would be preferable, but you should at least test the attractions.

backend/src/test/java/com/google/sps/TripCrudTest.java
mecolmg, line 87: Same here with testing the entire expected JSON or at least the attractions.

backend/src/test/java/com/google/sps/TripCrudTest.java
mecolmg, line 57: nit: for readability separate your actions and expectations by a newline here and below.

```suggestion
    Entity userEntity = UserCrud.createUser(EMAIL);
    Entity tripEntity = TripCrud.createTrip(EMAIL, TRIP_DATA);
    Entity readTripEntity = TripCrud.readTrip(Long.toString(tripEntity.getKey().getId()));

    assertEquals(tripEntity, readTripEntity);
```

backend/src/test/java/com/google/sps/TripCrudTest.java
```diff
@@ -0,0 +1,132 @@
...
+    Entity tripEntity = TripCrud.createTrip(EMAIL, TRIP_DATA);
+    TripCrud.updateTrip(Long.toString(tripEntity.getKey().getId()), TRIP_DATA_2);
+    Entity tripFound = TripCrud.readTrip(Long.toString(tripEntity.getKey().getId()));
+    assertEquals("\"My Awesome Milan Trip\"", (String) tripFound.getProperty("tripName"));
+  }
+
+  @Test
+  public void updateTrip_noTripFound() {
+    Entity actual = TripCrud.readTrip(INVALID_TRIP_ID);
+    assertEquals(null, actual);
```
mecolmg, line 102: nit: assertNull

backend/src/test/java/com/google/sps/TripCrudTest.java
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  @Test(expected = Exception.class)
```
mecolmg, line 105: nit: Is there a more specific exception you can verify throws than `Exception`? The reason being, if something else breaks in createTrip, this test will still pass. Same below

backend/src/main/java/com/google/sps/TripCrud.java
Cindyzhang977: We changed the way the json file was represented a while ago, so we had to update this to match.

backend/src/main/java/com/google/sps/TripCrud.java
Cindyzhang977: This is added to send the json to the front end. The tripId is used to get the trip from the database and if it's null it indicates the trip must be created.

backend/src/main/java/com/google/sps/TripCrud.java
Cindyzhang977: `tripEntity` was updated in `TripCrud.addTripId` so we should read it again from the database to get the most up to date version.

backend/src/test/java/com/google/sps/TripCrudTest.java
RCAVelez: done

backend/src/test/java/com/google/sps/TripCrudTest.java
RCAVelez: done

backend/src/test/java/com/google/sps/TripCrudTest.java
RCAVelez: I removed the readTrip_returnsEntityForTrip test

backend/src/test/java/com/google/sps/TripCrudTest.java
RCAVelez: done, I added a space between the assert and the logic above it on all tests

backend/src/test/java/com/google/sps/TripCrudTest.java
RCAVelez: done. All expected outcomes now have the a specific exception. 
These type of tests seem hacky. Are these fine?

backend/src/main/java/com/google/sps/TripCrud.java
RCAVelez: done, it's private

backend/src/main/java/com/google/sps/TripCrud.java
RCAVelez: readTrip, updateTrip, and toEntity all now take in a tripId of type Long. The servlets and tests have now been changed to reflect that change in type.

backend/src/main/java/com/google/sps/TripCrud.java
RCAVelez: Done! updateTrip now takes in a tripId of type Long. The servlets and tests have now been changed to reflect that change in type.

backend/src/test/java/com/google/sps/TripCrudTest.java
RCAVelez: We are having an issue with the toEntity methods where the additional quote is added. In order to pass this test we have to add an additional quote. Then in the frontend we have to remove that additional quote.

backend/src/test/java/com/google/sps/TripCrudTest.java
RCAVelez: We are having an issue with the toJson method where the additional quote is added. In order to pass this test we have to add an additional quote. Then in the frontend we have to remove that additional quote.

backend/src/test/java/com/google/sps/TripCrudTest.java
RCAVelez: yes we can

backend/src/test/java/com/google/sps/TripCrudTest.java
RCAVelez: can do

backend/src/main/java/com/google/sps/TripCrud.java
RCAVelez: Could we convert this to an issue and resolve it Monday?

backend/src/main/java/com/google/sps/TripCrud.java
RCAVelez: should be fixed now

backend/src/main/java/com/google/sps/TripCrud.java
mecolmg: Sure.

backend/src/test/java/com/google/sps/TripCrudTest.java
mecolmg: I guess that's my point. That means something is wrong, that needs to be fixed. The current workaround seems very hacky.

mecolmg: This can be converted to an issue and solved in a later PR

mecolmg: I don't see any changes though

backend/src/test/java/com/google/sps/TripCrudTest.java
mecolmg: I also don't see this in the changes.

backend/src/test/java/com/google/sps/TripCrudTest.java
RCAVelez: done

backend/src/test/java/com/google/sps/TripCrudTest.java
RCAVelez: done, I created a test for every property of attraction, and I was able to remove the extra quotes

backend/src/test/java/com/google/sps/TripCrudTest.java
RCAVelez: I was able to remove the extra quotes

backend/src/main/java/com/google/sps/TripCrud.java
```diff
@@ -68,18 +69,23 @@ private static void setProperties(Entity tripEntity, String tripData) {
...
-    tripEntity.setProperty("searchText", jsonObject.get("searchText").toString());
-    tripEntity.setProperty("tripName", jsonObject.get("tripName").toString());
-    tripEntity.setProperty("centerLocation", jsonObject.get("centerLocation").toString());
+    tripEntity.setProperty("searchText", jsonObject.get("searchText").getAsString());
+    tripEntity.setProperty("tripName", jsonObject.get("tripName").getAsString());
+
+    Integer centerLat = jsonObject.get("centerLat").getAsInt();
+    Integer centerLng = jsonObject.get("centerLng").getAsInt();
+    tripEntity.setProperty("centerLng", centerLng);
+    tripEntity.setProperty("centerLat", centerLat);
```
mecolmg, line 78: optional: You could make the centerLocation an `EmbededEntity` that has `lat` and `lng`.

backend/src/main/java/com/google/sps/TripCrud.java
```diff
@@ -68,18 +69,23 @@ private static void setProperties(Entity tripEntity, String tripData) {
     JsonElement jsonElement = parser.parse(tripData);
     JsonObject jsonObject = jsonElement.getAsJsonObject();
     tripEntity.setProperty("isOptimized", jsonObject.get("isOptimized").getAsBoolean());
-    tripEntity.setProperty("searchText", jsonObject.get("searchText").toString());
-    tripEntity.setProperty("tripName", jsonObject.get("tripName").toString());
-    tripEntity.setProperty("centerLocation", jsonObject.get("centerLocation").toString());
+    tripEntity.setProperty("searchText", jsonObject.get("searchText").getAsString());
```
mecolmg, line 72: Ah, nice catch! I'm glad you caught this. FYI @Cindyzhang977 you may need to adjust for this fix in your dependent PRs

backend/src/test/java/com/google/sps/TripCrudTest.java
```diff
@@ -72,7 +76,80 @@ public void readTrip_noTripFound() {
   public void toEntity_returnsTripEntityFromJsonMatchingTripName() {
     Entity tripEntityConverted = TripCrud.toEntity(TRIP_DATA, null, null);
 
-    assertEquals("\"My Milan Trip\"", (String) tripEntityConverted.getProperty("tripName"));
+    assertEquals("My Milan Trip", (String) tripEntityConverted.getProperty("tripName"));
```
mecolmg, line 79: optional: All these assertions could be done in a single test. It's fine to have multiple assertions in a single test, but they should generally be asserting the same thing. Here, you're verifying that toEntity converts from a json object to an entity.

